PR #23, opened 2018-04-27
CRUD buttons are visible only for logged in users

ValentinTyhonov:
+ removed annoyed error at the start
+ CRUD buttons are visibled only for logged in user's subscriptions
+ AddSubscription and BulkDelete buttons are visibled only for logged in users
+ added username column to subscriptions table, visibled if backend is secured
+ added checking is backend secured or not
+ fixed logout (session is removed from db)
+ fixed searching (subscription table)
+ authorized user can remove or modify only own subscriptions (checked on backend)


xmairaa:
can you show 5 min demo on this?

ValentinTyhonov:
Yes. Implementation is not finished, because I have few questions.. But part of this task is done,

src/main/java/com/ericsson/ei/frontend/EIRequestsController.java
```diff
@@ -93,7 +93,7 @@ public void setUseSecureHttp(boolean useSecureHttp) {
      *
      */
     @CrossOrigin
-    @RequestMapping(value = "/auth/*", method = RequestMethod.GET)
+    @RequestMapping(value = {"/auth/login", "/auth/logout"}, method = RequestMethod.GET)
     public ResponseEntity<String> getAuthRequests(Model model, HttpServletRequest request) {
         String eiBackendAddressSuffix = request.getServletPath();
```
tobiasake: It would be good to log the Login/Logout request, so we can debug the traffic. Maybe not the content of the request(with user/pass), but that a login request has been made from EI Fronted UI towards and also for the response.

ValentinTyhonov: For login/logout request we have next log (same as for other requests):

LOG.info("Got HTTP Request with method GET.\nUrlSuffix: " + eiBackendAddressSuffix +  "\nForwarding Request to EI Backend with url: " + newRequestUrl);

Is it enough?

PR #25, opened 2018-04-27
ADD: handling several EI instances

ned29:


ned29:
Need to improve frontend design, but backend works good

src/main/java/com/ericsson/ei/frontend/WebController.java
```diff
@@ -16,137 +16,107 @@
...
-    // Backend host and port (Getter & Setters), application.properties ->
-    // greeting.xxx
-    public String getFrontendServiceHost() {
-        return frontendServiceHost;
-    }
-
-    public void setFrontendServiceHost(String frontendServiceHost) {
-        this.frontendServiceHost = frontendServiceHost;
+    @RequestMapping("/register.html")
+    public String register(Model model) {
```
ValentinTyhonov: Remove this method. It was removed in my PR earlier.

src/main/java/com/ericsson/ei/frontend/WebController.java
```diff
@@ -16,137 +16,107 @@
...
+        return "register";
     }
 
-    public int getFrontendServicePort() {
-        return frontendServicePort;
+    @RequestMapping("/forgot-password.html")
+    public String forgotPassword(Model model) {
```
ValentinTyhonov: it should be removed too

pom.xml
```diff
@@ -54,6 +54,18 @@
...
+			<artifactId>lombok</artifactId>
+			<version>1.16.20</version>
+			<scope>provided</scope>
+		</dependency>
+		<dependency>
+			<groupId>com.google.code.gson</groupId>
+			<artifactId>gson</artifactId>
+			<version>2.8.4</version>
+		</dependency>
+
```
tobiasake: Has these new components been Foss checked/reported?

src/main/java/com/ericsson/ei/frontend/BackEndInformationControllerImpl.java
```diff
@@ -0,0 +1,89 @@
...
+
+    public ResponseEntity<String> getInstances(Model model) {
+        return new ResponseEntity<>(utils.getInstances().toString(), HttpStatus.OK);
+    }
+
+    public ResponseEntity<String> switchBackEndInstance(Model model, HttpServletRequest request) {
+        try {
+            String body = request.getReader().lines().collect(Collectors.joining(System.lineSeparator()));
+            utils.setInstances(new JsonParser().parse(body).getAsJsonArray());
+            utils.writeIntoFile();
```
tobiasake: Don't we have main EI backend in each Eiffel domain? Can't we store this info in main EI Backend MongoDb?

src/main/resources/static/js/switch-instances.js
```diff
@@ -0,0 +1,79 @@
...
+	    var json = JSON.parse(ko.toJSON(instances));
+	    for(var i = 0; i < json.length; i++){
+	        var obj = json[i];
+	        if(obj.active == true){
+	            count++;
+	        }
+	    }
+	    if(count == 1){
+		    $.ajax({
+			    url: "/switch-backend",
```
tobiasake: Do we need to have "frontendServiceUrl" value before "/switch-backend" suffix.
I guess Ajax presume the "/switch-backend" without url prefix should always connect to same EI Frontend(and EI backend) server as EI Frontend WebUI load web page from. Or?
EIBackendConnectionStatus need to have the "frontendServiceUrl" variable added to the url path in Ajax request. If don't have it, I guess it always check the connection status on main EI Backend instance.

ned29: fixed

src/main/java/com/ericsson/ei/frontend/BackEndInformationControllerImpl.java
ned29: it was Vali recommendation to store info in .json file

pom.xml
ned29: How can i checked/reported? working with knouckout and need gson

tobiasake: We have meeting on Wednesday about FOSSing new components. But if you want to do it directly you need to talk/chat with emichaf.

src/main/java/com/ericsson/ei/frontend/BackEndInformationControllerImpl.java
tobiasake: Ok

tobiasake: Which path are you storing to?
It can be permission problems depending on where and how EI-F is deployed.
If the path is configure by user in application.properties, then user can specify a path where EI-F has write permission.

ned29: You can specify path in application.properties file if path null by default will be "src/main/resources/EIBackendInstancesInformation.json"

ned29:
got two approvals, so merged

PR #28, opened 2018-05-18
EI Subscription REST POST supporting authentication headers

sericcson:


src/main/resources/application.properties
```diff
@@ -4,7 +4,7 @@ spring.jackson.default-property-inclusion=non_null
 spring.application.name=ei-frontend
   #${project.artifactId}
 
-server.port=8080
+server.port=8086
```
tobiasake: change back so 8080 is default port for webserver.

src/main/resources/application.properties
```diff
@@ -13,7 +13,7 @@ server.port=8080
 
 ######## EI Backend
 ei.frontendServiceHost=localhost
-ei.frontendServicePort=8080
+ei.frontendServicePort=8086
```
tobiasake: change back so 8080 is default port for webserver.

src/main/resources/static/resources/subscription_templates.js
```diff
@@ -16,6 +16,12 @@ var template_vars = {};
                         "formvalue" : ""
                     }
                 ],
+            "notificationMessageKeyValuesAuth" : [
```
tobiasake: I thought all subscription templates was moved to EI Backend since some months back. If you use  and update this template you will have to update EI Backend with same changes. There is RestApi to retrieve the Subscription template from EI Backend. I suggest to change so Subscription template content is read from EI Backend, so we use one and same subscription template in all places and we will then have only one subscription template file to keep updated.

src/main/resources/templates/subscription.html
```diff
@@ -169,10 +170,38 @@ <h3 class="modal-title text-center">Subscription Form</h3>
                                         class="glyphicon glyphicon-plus float-right"></i> Add Key/Value Pair
                                 </button>
                             </div>
-
+                        </div>                        
+                        
+                        <div class="form-group">
```
ValentinTyhonov: add here class pt-3 for example (like div class="pt-3 form-group"). It will add little bit more space between this and previous blocks and 'Authorization' labe; will not shift.

ValentinTyhonov:
Looks good

PR #30, opened 2018-05-21
Added frontendServiceUrl to each ajax request, where it was missed

ValentinTyhonov:


src/main/java/com/ericsson/ei/frontend/WebController.java
```diff
@@ -43,29 +43,21 @@
...
-            frontendServiceUrl = String.format("%s://%s:%d", httpMethod, frontendServiceHost, frontendServicePort);
-        }
-        model.addAttribute("frontendServiceUrl", frontendServiceUrl);  // inject in DOM for AJAX etc
+        model.addAttribute("frontendServiceUrl", getFrontendServiceUrl());  // inject in DOM for AJAX etc
         return "subscription";
     }
 
     @RequestMapping("/testRules.html")
     public String testRules(Model model) {
+        model.addAttribute("frontendServiceUrl", getFrontendServiceUrl());  // inject in DOM for AJAX etc
```
tobiasake: Is it required to have frontendServiceUrl add to every page?

ValentinTyhonov: If we want to have it in each ajax request then yes. in other case in js file we will have empty string when call frontendServiceUrl

tobiasake: Ok,, just checking so we use the variable on all page.

src/main/java/com/ericsson/ei/frontend/EIRequestsController.java
```diff
@@ -177,7 +177,7 @@ private String getEIBackendSubscriptionAddress() {
     private String getEIRequestURL(HttpServletRequest request) {
         String eiBackendAddressSuffix = request.getServletPath();
         String requestQuery = request.getQueryString();
-        String query = (requestQuery != null && !requestQuery.isEmpty()) ? "?" + requestQuery : "";
+        String query = (requestQuery != null && !requestQuery.isEmpty() && !requestQuery.startsWith("_")) ? "?" + requestQuery : "";
```
tobiasake: Why do we get these "?_2342342" stuff in the request? Shouldn't it be fixed in JS code where the request is made?

src/main/java/com/ericsson/ei/frontend/EIRequestsController.java
```diff
@@ -177,8 +182,13 @@ private String getEIBackendSubscriptionAddress() {
     private String getEIRequestURL(HttpServletRequest request) {
         String eiBackendAddressSuffix = request.getServletPath();
         String requestQuery = request.getQueryString();
-        String query = (requestQuery != null && !requestQuery.isEmpty() && !requestQuery.startsWith("_")) ? "?" + requestQuery : "";
-        String requestUrl = getEIBackendSubscriptionAddress() + eiBackendAddressSuffix + query;
+        String query = (requestQuery != null && !requestQuery.isEmpty()) ? "?" + requestQuery : "";
```
tobiasake: Move in String query line into the if statement where query variable is used.

tobiasake: Move String requestQuery line to if statement also.

PR #36, opened 2018-06-07
Create EI Frontend Test Framework

eroohen:
* Create EI Frontend Test Framework with Selenium test framework (using Firefox driver).
 * This test framework will be used for EI Frontend Functional Tests.
 * Support for Travis-CI
 * Sample test case


eroohen:
We did not think about BDD at the time we wrote this. However, after investigating this carefully the concepts of GUI testing and BDD seems not very compatible together (see [link](https://www.testingexcellence.com/selenium-and-cucumber-ui-automation-challenges/)) . In GUI testing we want a larger "journey" tests where in cucumber we want small isolated tests which is hard to make in a GUI environment without making it very slow, since we need to traverse the webpage to get to the necessary state.

ValentinTyhonov:
Yes, if backend is mocked right, then /auth/checkStatus request should return 200 and button should be green

src/functionaltest/java/com/ericsson/ei/frontend/TemplateTestCase.java
```diff
@@ -0,0 +1,28 @@
...
...ontype\":\"eiffel-intelligence\",\"created\":1524037895415,\"notificationMeta\":\"http://eiffel-jenkins1:8080/job/ei-artifact-triggered-job/build\",\"notificationType\":\"REST_POST\",\"restPostBodyMediaType\":\"application/x-www-form-urlencoded\",\"notificationMessageKeyValues\":[{\"formkey\":\"json\",\"formvalue\":\"{parameter: [{ name: 'jsonparams', value : to_string(@) }]}\"}],\"repeat\":false,\"requirements\":[{\"conditions\":[{\"jmespath\":\"gav.groupId=='com.othercompany.library'\"}]}],\"subscriptionName\":\"Subscription2\",\"_id\":{\"$oid\":\"5ad6f907c242af3f1469751e\"}},{\"aggregationtype\":\"eiffel-intelligence\",\"created\":1524223397628,\"notificationMeta\":\"http://<MyHost:port>/api/doit\",\"notificationType\":\"REST_POST\",\"restPostBodyMediaType\":\"application/json\",\"notificationMessageKeyValues\":[{\"formkey\":\"\",\"formvalue\":\"{mydata: [{ fullaggregation : to_string(@) }]}\"}],\"repeat\":false,\"requirements\":[{\"conditions\":[{\"jmespath\":\"submission.sourceChanges[?submitter.group == 'Team Gophers' && svnIdentifier==null]\"}]}],\"subscriptionName\":\"Subscription_Template_Rest_Post_Raw_Body_Json_Trigger\",\"_id\":{\"$oid\":\"5ad9cda5b715d336247e4980\"}}]";
```
ValentinTyhonov: can it be loaded from file?

e-edling-ericsson:
This pull request only contains a template for testing, not actual tests. But yes, some story should cover that the ""EI Backend Status" is getting updated correctly depending on the checkStatus-request response

src/main/java/com/ericsson/ei/frontend/WebController.java
```diff
@@ -43,35 +47,67 @@
 
     @RequestMapping("/")
     public String greeting(Model model) {
-        model.addAttribute("frontendServiceUrl", getFrontendServiceUrl());  // inject in DOM for AJAX etc
+        // inject
+        // in
+        // DOM
+        // for
+        // AJAX
+        // etc
```
ValentinTyhonov: why everywhere comments changed to one word per line?

src/functionaltest/java/com/ericsson/ei/frontend/TemplateTestCase.java
e-edling-ericsson: Absolutely, I didn´t know what way we prefered. There are both positives and negatives about both approaches(i prefer it in the actual tests as it becomes easier to get an overview of what the test is actually doing and returning - the drawback is the cluttering) but if you guys prefer from file i´ll change that.

src/main/java/com/ericsson/ei/frontend/WebController.java
e-edling-ericsson: I think the eclipse style formatter that we use does this. Should i "overrule" the formatter or leave this?

src/functionaltest/java/com/ericsson/ei/frontend/TemplateTestCase.java
ValentinTyhonov: My opinion is to save it in file. Then this data will be more readable and can be reused in other tests, if needed, without coping and pasting this big part of text. But you can also check it with other teammates.

src/main/java/com/ericsson/ei/frontend/WebController.java
ValentinTyhonov: Better do as it was before. As for me, I'm not sure if these comments are still needed there

src/functionaltest/java/com/ericsson/ei/frontend/TemplateTestCase.java
e-edling-ericsson: Absolutely, i guess its easier with these big responses. I´ll fix that!

src/main/java/com/ericsson/ei/frontend/WebController.java
e-edling-ericsson: I guess i´ll disable the formatter then as it does this automatically, i completely agree with you.

src/functionaltest/java/com/ericsson/ei/frontend/pageobjects/PageBaseClass.java
```diff
@@ -0,0 +1,65 @@
...
+import org.mockito.ArgumentMatcher;
+import org.mockito.Mockito;
+import org.openqa.selenium.JavascriptExecutor;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.support.PageFactory;
+import org.openqa.selenium.support.ui.ExpectedCondition;
+import org.openqa.selenium.support.ui.WebDriverWait;
+
+public class PageBaseClass {
+    // EIRequestsController mockEIRequestsController;
```
evasiba-ericsson: remove unused commented out code

src/functionaltest/java/com/ericsson/ei/frontend/pageobjects/PageBaseClass.java
```diff
@@ -0,0 +1,65 @@
...
+            WebDriver driver, String baseUrl) throws ClientProtocolException, IOException {
+        super();
+        this.mockedHttpClient = mockedHttpClient;
+        this.driver = driver;
+        this.baseUrl = baseUrl;
+        PageFactory.initElements(driver, this);
+
+        //Dummy response for all requests that happens before the actuall ones we want to test
+        CloseableHttpResponse response = this.createMockedHTTPResponse("{\"response\": dummy}", 200);
+        Mockito.doReturn(response).when(mockedHttpClient).execute(Mockito.argThat(request -> ((HttpRequestBase)request).getURI().toString().contains("checkStatus")));
```
evasiba-ericsson: row too long

pom.xml
```diff
@@ -1,115 +1,166 @@
...
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-all</artifactId>
+      <version>1.10.19</version>
+    </dependency>
+    <!-- https://mvnrepository.com/artifact/org.mock-server/mockserver-client-java -->
+    <dependency>
+      <groupId>org.seleniumhq.selenium</groupId>
+      <artifactId>selenium-firefox-driver</artifactId>
+      <version>3.12.0</version>
```
tobiasake: Which Firefox version corresponds this Firefox-driver version to? Do we have no possibility to test with a specific Firefox version?

src/functionaltest/java/com/ericsson/ei/frontend/TemplateTestCase.java
tobiasake: I prefer this to be in a file also.

src/functionaltest/java/com/ericsson/ei/frontend/pageobjects/TestRulesPage.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class TestRulesPage extends PageBaseClass {
+
+    public TestRulesPage(CloseableHttpClient mockedHttpClient,
+            WebDriver driver, String baseUrl) throws ClientProtocolException, IOException {
+        super(mockedHttpClient, driver, baseUrl);
+    }
+
+    public String getMainHeader() {
+        WebElement mainHeader = driver.findElement(By.xpath("/html/body/div[3]/div[1]/div[2]/div[1]/h1"));
```
tobiasake: How does this work? If I add a new div tag in the middle of this to html code,, will that break the tests?
Is it possible to connect this to an id parameter in the div tag instead.

pom.xml
tobiasake: Create a new story for running with different Firefox version and use Firefox driver and headless mode for the moment.

src/functionaltest/java/com/ericsson/ei/frontend/pageobjects/TestRulesPage.java
e-edling-ericsson: That is true. The specific element doesn't have ID or class identifier though, i guess we can add one instead of using xpath.

src/functionaltest/java/com/ericsson/ei/frontend/SeleniumBaseClass.java
```diff
@@ -0,0 +1,69 @@
...
+        webController.setFrontendServicePort(randomServerPort);
+        driver = SeleniumConfig.getFirefoxDriver();
+        baseUrl = SeleniumConfig.getBaseUrl(randomServerPort);
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        driver.quit();
+        String verificationErrorString = verificationErrors.toString();
+        if (!"".equals(verificationErrorString)) {
```
ValentinTyhonov: not a big problem, but maybe !verificationErrorString.equals("") ?

src/functionaltest/resources/responses/SubscriptionObjects.json
```diff
@@ -0,0 +1 @@
...y'"}]}],"subscriptionName":"Subscription1","_id":{"$oid":"5ad6f907c242af3f1469751d"}},{"aggregationtype":"eiffel-intelligence","created":1524037895415,"notificationMeta":"http://eiffel-jenkins1:8080/job/ei-artifact-triggered-job/build","notificationType":"REST_POST","restPostBodyMediaType":"application/x-www-form-urlencoded","notificationMessageKeyValues":[{"formkey":"json","formvalue":"{parameter: [{ name: 'jsonparams', value : to_string(@) }]}"}],"repeat":false,"requirements":[{"conditions":[{"jmespath":"gav.groupId=='com.othercompany.library'"}]}],"subscriptionName":"Subscription2","_id":{"$oid":"5ad6f907c242af3f1469751e"}},{"aggregationtype":"eiffel-intelligence","created":1524223397628,"notificationMeta":"http://<MyHost:port>/api/doit","notificationType":"REST_POST","restPostBodyMediaType":"application/json","notificationMessageKeyValues":[{"formkey":"","formvalue":"{mydata: [{ fullaggregation : to_string(@) }]}"}],"repeat":false,"requirements":[{"conditions":[{"jmespath":"submission.sourceChanges[?submitter.group == 'Team Gophers' && svnIdentifier==null]"}]}],"subscriptionName":"Subscription_Template_Rest_Post_Raw_Body_Json_Trigger","_id":{"$oid":"5ad9cda5b715d336247e4980"}}]
```
ValentinTyhonov: format it as pretty

src/functionaltest/java/com/ericsson/ei/frontend/SeleniumBaseClass.java
e-edling-ericsson: Agreed, looks better

pom.xml
```diff
@@ -1,114 +1,171 @@
...
+      <groupId>org.mock-server</groupId>
+      <artifactId>mockserver-netty</artifactId>
+      <version>5.3.0</version>
+      <scope>test</scope>
+    </dependency>
+	<!-- https://mvnrepository.com/artifact/org.json/json -->
+	  <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <version>20171018</version>
```
ValentinTyhonov: Why did you downlift version of org.json ? Are there some reasons for doing it?

PR #38, opened 2018-06-13
Added new info in EI Instance Info page and Remove gson version tag from pom file.

tobiasake:


tobiasake:
We have created story solving that issue, check the story for the solution. But I will update the default value suggestion.

PR #39, opened 2018-06-13
Fix properties

ned29:
fix issue when values in properties file are not defined or null


src/main/java/com/ericsson/ei/frontend/BackEndInformationController.java
```diff
@@ -16,35 +16,107 @@
...
+        try {
+            String body = request.getReader().lines().collect(Collectors.joining(System.lineSeparator()));
+            backEndInstancesUtils.setInstances(new JsonParser().parse(body).getAsJsonArray());
+            backEndInstancesUtils.writeIntoFile();
+            for (BackEndInformation backEndInformation : backEndInstancesUtils.getInformation()) {
+                if (backEndInformation.isActive()) {
+                    backEndInstancesUtils.setBackEndProperties(backEndInformation);
+                }
+            }
+            return new ResponseEntity<>(HttpStatus.OK);
```
tobiasake: Don't we need to JSON Content Header here in the response. When I looked at AJAX call earlier, it expects the JSON content header.
This comment applies to all Responses here.

src/main/java/com/ericsson/ei/frontend/BackEndInformationController.java
```diff
@@ -16,35 +16,107 @@
...
+    public ResponseEntity<String> getInstances(Model model) {
+        return new ResponseEntity<>(backEndInstancesUtils.getInstances().toString(), HttpStatus.OK);
+    }
 
-    @CrossOrigin
     @RequestMapping(value = "/switch-backend", method = RequestMethod.POST)
-    ResponseEntity<String> switchBackEndInstance(Model model, HttpServletRequest request);
+    public ResponseEntity<String> switchBackEndInstance(Model model, HttpServletRequest request) {
+        try {
+            String body = request.getReader().lines().collect(Collectors.joining(System.lineSeparator()));
```
tobiasake: What does body contain? EI Instance Url?

src/main/java/com/ericsson/ei/frontend/BackEndInformationController.java
```diff
@@ -16,35 +16,107 @@
...
-    @CrossOrigin
-    @RequestMapping(value = "/switch-backendByMainPage", method = RequestMethod.POST)
-    ResponseEntity<String> switchBackEndInstanceByMainPage(Model model, HttpServletRequest request);
-
-    @CrossOrigin
     @RequestMapping(value = "/switch-backend", method = RequestMethod.DELETE)
-    ResponseEntity<String> deleteBackEndInstance(Model model, HttpServletRequest request);
+    public ResponseEntity<String> deleteBackEndInstance(Model model, HttpServletRequest request) {
+        try {
+            String body = request.getReader().lines().collect(Collectors.joining(System.lineSeparator()));
```
tobiasake: I think you can make clearer here also, what data in Body you use or what body contain.

tobiasake: Or document what it contains.

src/main/java/com/ericsson/ei/frontend/BackEndInformationController.java
```diff
@@ -16,35 +16,107 @@
...
+            } else {
+                return new ResponseEntity<>("Instance already exist", HttpStatus.BAD_REQUEST);
+            }
+        } catch (Exception e) {
+            return new ResponseEntity<>("Internal error", HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    @RequestMapping(value = "/switchBackend", method = RequestMethod.POST)
+    public ResponseEntity<String> switchBackEndInstanceByMainPage(Model model, HttpServletRequest request) {
```
tobiasake: What do you mean "ByMainPage"? ,, switch is only steering to which address to do RestApi calls to.

tobiasake: "frontendServiceUrl" and same variable in Bridge should be updated with new url when switch is made.

src/main/resources/static/js/main.js
```diff
@@ -113,6 +113,7 @@ jQuery(document).ready(function() {
     }
 
     function viewModel(data) {
+    console.log(data);
```
tobiasake: I guess this is just for debugging for the moment.

src/main/java/com/ericsson/ei/frontend/BackEndInformationController.java
ned29: We have two chooses in switching backends first from drop-list(current) and second from page

src/main/resources/static/js/main.js
ned29: removed

src/main/java/com/ericsson/ei/frontend/BackEndInformationController.java
ned29: renamed

src/main/java/com/ericsson/ei/frontend/BackEndInformationController.java
ned29: don't clearly understand what do you mean?

tobiasake: From the error in the bug that was reported,, what I could find about that was some info about that the Header ContentType is missing and might be the root cause to the error.
So this might solve the error.

Ajax Call (JS) expect HEADER ContentType JSON,,, so ResponseEntity need to have something added:
HttpHeaders headers = new HttpHeaders();
        headers.setContentType(MediaType.APPLICATION_JSON);
        return new ResponseEntity<>(responseBody, headers, HttpStatus.valueOf(statusCode));

ned29: ok, thanks will fix it

ned29: Added

src/main/java/com/ericsson/ei/frontend/BackEndInformationController.java
```diff
@@ -16,35 +16,112 @@
...
+                info.add(backEndInformation);
+            }
+            backEndInstancesUtils.setInformation(info);
+            JsonArray result = (JsonArray) new Gson().toJsonTree(backEndInstancesUtils.getInformation(), new TypeToken<List<BackEndInformation>>() {
+            }.getType());
+            backEndInstancesUtils.setInstances(result);
+            backEndInstancesUtils.writeIntoFile();
+            return new ResponseEntity<>(HttpStatus.OK);
+        } catch (Exception e) {
+            return new ResponseEntity<>("Internal error", getHeaders(), HttpStatus.INTERNAL_SERVER_ERROR);
```
tobiasake: Include Exception message into the response message. Include more info of what failed(or what possibly failed.)
"Internal error" doesn't tell user/admin that much.

tobiasake: getHeaders(),, should be added to all Entrypoints that return Responses where JS Ajax requests expects JSON as ContentType Header.

tobiasake: Check the JS code and Ajax request and what the Ajax request expects from these entry points.

src/main/java/com/ericsson/ei/frontend/BackEndInformationController.java
```diff
@@ -16,35 +16,112 @@
...
 import org.springframework.ui.Model;
-import org.springframework.web.bind.annotation.CrossOrigin;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
-import org.springframework.web.bind.annotation.RestController;
 
 import javax.servlet.http.HttpServletRequest;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
```
tobiasake: Why remove the RAML based interface? Architects has decided to use RAML.

ned29: This interface wasn't base on RAML and don't use raml in frontend

src/main/java/com/ericsson/ei/frontend/BackEndInformationController.java
ned29: added

src/main/resources/application.properties
```diff
@@ -15,10 +15,11 @@ server.port=8080
 ei.frontendServiceHost=localhost
 ei.frontendServicePort=8080
 ei.frontendContextPath=
+ei.useSecureHttpForFrontend=false
 ei.backendServerHost=localhost
 ei.backendServerPort=8090
 ei.backendContextPath=
-ei.useSecureHttp=false
+ei.useSecureHttpForBackend=false
```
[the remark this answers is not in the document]
tobiasake: I saw now that you already have a checkbox for HTTPS in the admin page for Switching EI Backend instances.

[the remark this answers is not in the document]
tobiasake: I think you should add Mouse hover over info box that show example of how the Url should be specified, How context path should be specified, etc. on the Admin Swith EI Instance Page.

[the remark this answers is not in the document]
ned29: it's possible configure https or http from switching but if we want to configure backend using applications,properties file. this field isn't mandatory by default will be false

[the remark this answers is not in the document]
tobiasake: ok. I didn't get the full picture how it was today in this review view.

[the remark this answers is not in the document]
ned29: Mouse hover means example of gotten url?

src/test/java/com/ericsson/ei/frontend/BackendInformationControllerTest.java
```diff
@@ -86,7 +86,7 @@ public void testAddInstance() throws Exception {
         mockMvc.perform(MockMvcRequestBuilders.post("/add-instances")
             .accept(MediaType.APPLICATION_JSON_VALUE)
```
tobiasake: You can use real HTTP Request library also,,, depends how near real environment we want to test. Both are ok to use.

src/main/java/com/ericsson/ei/frontend/WebController.java
```diff
@@ -42,6 +42,9 @@
     @Value("${ei.eiffelDocumentationUrls}")
     private String eiffelDocumentationUrls;
 
+    @Value("${ei.useSecureHttpForFrontend}")
+    private boolean httpsForFrontend;
```
tobiasake: Why not name it with same name? e.g "useSecureHttpFrontend"

ned29: I separated https for frontend and backend because they use the same properties for this configuration and it's wrong assuming.

src/main/resources/application.properties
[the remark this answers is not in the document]
tobiasake: Mouse hover means when you hover the mouse over the Text Box field, a small Help Popup window shows up. In the Help popup window show an example of ei backend url , so user know in what format of url to specify in the url text box. User might wonder "should I specify url with http//:<address>:<port> or just "<address>:<port>" etc.

src/main/java/com/ericsson/ei/frontend/WebController.java
tobiasake: yes, but I mean you cna use same name in prpoperties and code.
@Value("${ei.useSecureHttpFrontend}")
private boolean useSecureHttpFrontend;

And for backend:
@Value("${ei.useSecureHttpBackend}")
private boolean useSecureHttpBackend;

ned29: renamed

src/main/resources/application.properties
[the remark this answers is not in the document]
ned29: added

PR #41, opened 2018-06-19
Fix serialization for gson

ned29:


src/main/java/com/ericsson/ei/frontend/model/BackEndInformation.java
```diff
@@ -41,6 +42,7 @@
 
     @Value("${ei.useSecureHttpBackend:#{false}}")
     @JsonProperty("https")
+    @SerializedName("https")
```
ValentinTyhonov: good, but why only for one field?

tobiasake: Why do we need this?

ned29: @JsonProperty("https") - this for jackson
@SerializedName("https") - for gson

because only in this case field name and serialized name different

PR #42, opened 2018-06-20
Updated regarding to updates in REST API

ValentinTyhonov:
* Updated regarding to updates in REST API
* Removed useless print in test


src/main/resources/static/js/subscription.js
```diff
@@ -452,24 +452,26 @@ jQuery(document).ready(function() {
...
+                 content: 'Are you sure you want to delete these subscriptions?<pre>' + subscriptionsToDeleteString,
                  buttons: {
                      confirm: function () {
                     	 var ajaxHttpSender = new AjaxHttpSender();
-                    	 for (i=0; i < subScriptionsToDelete.length; i++){
-                    		 ajaxHttpSender.sendAjax(frontendServiceUrl + "/subscriptions/"+subScriptionsToDelete[i], "DELETE", null, callback);
-                    	 }
+                    	 // replace all /n with comma
+                    	 subscriptionsToDeleteString = subscriptionsToDeleteString.replace(new RegExp('\n', 'g'), ',').slice(0, -1);
+                    	 console.log(subscriptionsToDeleteString);
```
tobiasake: Is this log for debugging? or shall i be kept? If it is going to be kept a message before the list would be good to have:
"Sucsctions to be deleted: " + subscriptionsToDeleteString.

src/main/resources/static/js/subscription.js
```diff
@@ -511,18 +513,20 @@ jQuery(document).ready(function() {
...
-                    $.jGrowl("Failed to create Subscription: " + subscriptionJson.subscriptionName + " Error: " + XMLHttpRequest.responseText, {
-                        sticky : false,
-                        theme : 'Error'
-                    });
+                  reload_table();
+                  $.jGrowl("Failed to create next Subscriptions", {sticky: false, theme: 'Error'});
+                  var responseJSON = JSON.parse(XMLHttpRequest.responseText);
+                  for (var i = 0; i < responseJSON.length; i++) {
+                    $.jGrowl(responseJSON[i].subscription + " :: " + responseJSON[i].reason, {sticky: true, theme: 'Error'});
+                  }
```
tobiasake: How will this work? If we have 30 subscriptions failure,, will the whole frontend  GUI be override of alot of Error popup windows. That will not fit a normal web browser window. We maybe should log all popup error windows that appers to these drop down menu lists(not used today) in top bar.

src/main/resources/static/js/subscription.js
ValentinTyhonov: Sorry. Forgot to remove.

src/main/resources/static/js/subscription.js
tobiasake: Will be solved in a coming story.

PR #46, opened 2018-07-02
Functional Tests Subscription

sericcson:


src/functionaltest/java/com/ericsson/ei/frontend/SubscriptionHandlingFunctionality.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Test
+    public void testSubscription() throws Exception {
+
+        // Open index page
+        IndexPage indexPageObject = new IndexPage(mockedHttpClient, driver, baseUrl);
+        indexPageObject.loadPage();
+
+        // Click on Subscription Handling page button and verify that it is open
+        String headerPath = "//div[@class='container pull-left']//h1";
```
e-edling-ericsson: I think it's better to use "find by ID" instead of xpath. If a minor thing is changed in the DOM structure this would more easily break than using ID. If an ID is not present i would suggest to add it to the html code :) This applies to all xpaths you are using.

src/functionaltest/java/com/ericsson/ei/frontend/pageobjects/IndexPage.java
```diff
@@ -37,6 +39,13 @@ public TestRulesPage clickTestRulesPage() throws ClientProtocolException, IOExce
 
         return testRulesPage;
     }
+    
+    public SubscriptionPage clickSubscriptionPage() throws ClientProtocolException, IOException {
+      WebElement subscriptionBtn = driver.findElement(By.id("subscriptionBtn"));
```
e-edling-ericsson: Timing issue will occur if the button isnt yet loaded into the html code. You would need a "wait until presence of subscriptionBtn" i think. This is applied to every time you click something and something else has to appear on the website. It's asynchronous so you cant expect it to be there at the moment you click, i saw this in a few places more in the code.

src/functionaltest/java/com/ericsson/ei/frontend/SubscriptionHandlingFunctionality.java
```diff
@@ -0,0 +1,138 @@
...
+        // Click "Add Subscription" button and verify that "Subscription Form" is open
+         subscriptionPage.clickAddSubscription();
+         String xPath = "//*[@id='formHeader']";
+         new WebDriverWait(driver, 10).until((webdriver) -> (subscriptionPage.presenceOfHeader(xPath)));
+    
+         // On subscription form, select the template as "Mail Trigger" and verify
+         String tempPath = "//select[contains(@title,'Choose a Subscription Template')]";
+         String tempMail = "Mail Trigger";
+         subscriptionPage.selectDropdown(tempPath, tempMail);
+         TimeUnit.SECONDS.sleep(2);
```
e-edling-ericsson: Do you really need a sleep here? If you are waiting for an item in the dropdown to appear, there is a function called 
ExpectedConditions.elementToBeClickable(By.id)

I think its the same with the rest sleeps. They should be possible to be replaced by ExpectedConditions

src/functionaltest/java/com/ericsson/ei/frontend/pageobjects/IndexPage.java
```diff
@@ -56,4 +66,5 @@ public Object presenceOfReloadButton() {
             return false;
         }
     }
```
e-edling-ericsson: This can be replaced by the ExpectedConditions i refered to before. I will change this in my code aswell :)

src/functionaltest/java/com/ericsson/ei/frontend/SubscriptionHandlingFunctionality.java
```diff
@@ -0,0 +1,138 @@
...
+        // // Press "Reload" button and verify that two subscriptions with names "Subscription1" and "Subscription2" are present
+         String response = this.getJSONStringFromFile(RELOADTESTFILEPATH);
+         subscriptionPage.clickReload(response);
+         assert (driver.getPageSource().contains("Subscription1"));
+         assert (driver.getPageSource().contains("Subscription2"));
+        
+        // Click "Add Subscription" button and verify that "Subscription Form" is open
+         subscriptionPage.clickAddSubscription();
+         String xPath = "//*[@id='formHeader']";
+         new WebDriverWait(driver, 10).until((webdriver) -> (subscriptionPage.presenceOfHeader(xPath)));
```
e-edling-ericsson: In my mind, i think it looks better if you move all the waits into the corresponding page interaction method. For example
before i click a button i must wait for it to be clickable.

I will do this in my code aswell :)

pom.xml
```diff
@@ -5,6 +5,7 @@
 
   <groupId>com.github.ericsson</groupId>
   <artifactId>eiffel-intelligence-frontend</artifactId>
+
```
a-breid-ericsson: Should it be an empty line here?

src/functionaltest/java/com/ericsson/ei/frontend/SeleniumBaseClass.java
```diff
@@ -59,6 +65,7 @@ public void setUp() throws Exception {
 
     @After
     public void tearDown() throws Exception {
+
```
a-breid-ericsson: First line in method is empty.

src/functionaltest/java/com/ericsson/ei/frontend/SeleniumBaseClass.java
```diff
@@ -43,6 +48,7 @@
     @Autowired
     WebController webController;
 
+
```
a-breid-ericsson: 2 empty lines instead of just 1

src/functionaltest/java/com/ericsson/ei/frontend/SubscriptionHandlingFunctionality.java
```diff
@@ -0,0 +1,138 @@
...
+        // Verify that "get template" button works
+        String mockedTemplateResponse = this.getJSONStringFromFile(SUBSCRIPTIONTEMPLATEFILEPATH);
+        new WebDriverWait(driver, 10).until((webdriver) -> subscriptionPage.presenceOfClickGetTemplateButton());
+        subscriptionPage.clickDownloadGetTemplate(mockedTemplateResponse);
+        new WebDriverWait(driver, 10).until((webdriver) -> Files.exists(Paths.get(DOWNLOADEDTEMPLATEFILEPATH)));
+        String getSubscriptionsTemplate = this.getJSONStringFromFile(DOWNLOADEDTEMPLATEFILEPATH);
+        assertEquals(mockedTemplateResponse, getSubscriptionsTemplate);
+        
+        // Upload a subscription, name as "Subscription_uploaded" with "Upload SUbscriptions" button and verify
+        String uploadFilePath = new File("src/test/resources/Subscription_upload.txt").getAbsolutePath();
```
a-breid-ericsson: Should this not be a file in functionaltest/resources?

src/functionaltest/resources/responses/Subscription_upload.txt
```diff
@@ -0,0 +1,39 @@
+{
```
a-breid-ericsson: Duplicate file?

src/main/java/com/ericsson/ei/frontend/WebController.java
```diff
@@ -58,18 +59,21 @@ public String greeting(Model model) {
 
     @RequestMapping("/subscriptionpage.html")
     public String subscription(Model model) {
+
```
a-breid-ericsson: Should functions start with an empty line?

src/test/resources/Subscription_upload.txt
```diff
@@ -0,0 +1,39 @@
+{
```
a-breid-ericsson: Should this be a .txt file and not json?

src/functionaltest/java/com/ericsson/ei/frontend/SubscriptionHandlingFunctionality.java
sericcson: Actually, in the current code state, I have option to use either absolute or xpath... so I preffered xpath for the obvious reasons. However, "find by ID" is a better one but as you mentioned html code need to added for ID.

src/functionaltest/java/com/ericsson/ei/frontend/SubscriptionHandlingFunctionality.java
```diff
@@ -0,0 +1,128 @@
...
+import java.nio.file.Paths;
+import java.util.concurrent.TimeUnit;
+
+import com.ericsson.ei.config.SeleniumConfig;
+import com.ericsson.ei.frontend.pageobjects.IndexPage;
+import com.ericsson.ei.frontend.pageobjects.SubscriptionPage;
+
+public class SubscriptionHandlingFunctionality extends SeleniumBaseClass {
+
+    private static final String DOWNLOADEDTEMPLATEFILEPATH = String.join(File.separator,
```
e-pettersson-ericsson: Using _ between caps in variables make them a lot easier to read at first glance, like: DOWNLOADED_TEMPLATE_FILE_PATH

src/functionaltest/java/com/ericsson/ei/frontend/SubscriptionHandlingFunctionality.java
```diff
@@ -0,0 +1,128 @@
...
+        subscriptionPage.clickUploadSubscriptionFunctionality(DOWNLOADEDTEMPLATEFILEPATH, mockedUploadResponse);
+        new WebDriverWait(driver, 10).until((webdriver) -> (driver.getPageSource().contains("Subscription_uploaded")));
+
+        // Click "Add Subscription" button and verify that "Subscription Form" is open
+        subscriptionPage.clickAddSubscription();
+        String xPath = "//*[@id='formHeader']";
+        new WebDriverWait(driver, 10).until((webdriver) -> (subscriptionPage.presenceOfHeader(xPath)));
+
+        // On subscription form, select the template as "Mail Trigger" and verify
+        // On subscription form, select the template as "Mail Trigger" and verify
```
e-pettersson-ericsson: Duplicate comments here.

src/functionaltest/resources/responses/Subscription_upload.txt
sericcson: necessary changes are made

src/functionaltest/java/com/ericsson/ei/frontend/SubscriptionHandlingFunctionality.java
sericcson: true..

src/functionaltest/java/com/ericsson/ei/frontend/pageobjects/SubscriptionPage.java
```diff
@@ -0,0 +1,149 @@
...
+        CloseableHttpResponse response = this.createMockedHTTPResponse(responseData, 200);
+        Mockito.doReturn(response).when(mockedHttpClient).execute(Mockito.argThat(
+                request -> ((HttpRequestBase) request).getURI().toString().contains("/subscriptionsTemplate")));
+        WebElement getTemplateButton = driver.findElement(By.className("get_subscription_template"));
+        getTemplateButton.click();
+    }
+    
+    public void clickViewBtn() {
+        WebElement viewBtn = wait.until(ExpectedConditions.elementToBeClickable(By.xpath("//button[contains(text(),'View')]")));
+        viewBtn.click();
```
tobiasake: Will this xpath lookup have impact if html tag order in code changes?
I thinks its better to lookup html tags/DOM objects by the Id tag.

src/main/resources/application.properties
```diff
@@ -4,7 +4,7 @@ spring.jackson.default-property-inclusion=non_null
 spring.application.name=ei-frontend
   #${project.artifactId}
 
-server.port=8080
+server.port=8086
```
tobiasake: Change back to the default 8080 port.

tobiasake:
Does these tests test both when LDAP is disabled and when LDAP is enabled?
According to the user story tests should test both when LDAP is disabled and LDAP is enabled.

src/functionaltest/java/com/ericsson/ei/frontend/pageobjects/SubscriptionPage.java
sericcson: Well, it is independent of path... xpth is not absolute path so it need not be changed with change in order.
You may notice that mostly the "ids" are used for localization but xpath is used when it is convenient.

src/main/resources/application.properties
sericcson: OK

src/functionaltest/java/com/ericsson/ei/frontend/pageobjects/SubscriptionPage.java
tobiasake: Ok. I think we should rely on ids where it is possible, so test wont't break if some new html tags is added. But if this is not impacted by that, then I think this is a OK solution.

PR #50, opened 2018-07-13
Add firefox bin installation for firefoxdriver

e-edling-ericsson:
- Firefox could not run on Linux where firefox did not exist,
  this fix makes it possible for firefox to download binaries and
  install them for firefoxdriver.
- Fix minor issue with mockservers teardown.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,6 +106,11 @@
       <version>5.4.1</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+            <groupId>org.rauschig</groupId>
+            <artifactId>jarchivelib</artifactId>
+            <version>0.7.1</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/pom.xml
+++ b/pom.xml
@@ -107,9 +107,9 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-            <groupId>org.rauschig</groupId>
-            <artifactId>jarchivelib</artifactId>
-            <version>0.7.1</version>
+      <groupId>org.rauschig</groupId>
+      <artifactId>jarchivelib</artifactId>
+      <version>0.7.1</version>
     </dependency>
   </dependencies>
 
--- a/src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
+++ b/src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
@@ -95,7 +95,7 @@ public class SeleniumConfig {
                 LOGGER.error("Failed to load properties");
                 return false;
             } else {
-                LOGGER.info("Properties have been loaded.");
+                LOGGER.debug("Properties have been loaded.");
                 return true;
             }
         } catch (NumberFormatException e) {
--- a/src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
+++ b/src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
@@ -11,6 +11,7 @@ import org.openqa.selenium.firefox.FirefoxProfile;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 
 import java.util.Properties;
@@ -76,8 +77,7 @@ public class SeleniumConfig {
     }
 
     private static FirefoxBinary installFirefoxBinary() {
-        String[] firefoxBZip2FileUrlLinuxBits = firefoxBZip2FileUrlLinux.split("/");
-        String firefoxBZip2FileNameLinux = firefoxBZip2FileUrlLinuxBits[firefoxBZip2FileUrlLinuxBits.length-1];
+        String firefoxBZip2FileNameLinux = FilenameUtils.getName(firefoxBZip2FileUrlLinux);
 
         String firefoxBZip2FilePath = String.join(
                 File.separator, tempDownloadDirectory.getPath(), firefoxBZip2FileNameLinux);
--- a/src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
+++ b/src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
@@ -1,33 +1,65 @@
 package com.ericsson.ei.config;
 
 import com.ericsson.ei.frontend.exception.OSNotSupportedException;
-
+import com.ericsson.ei.frontend.exception.PropertiesNotLoadedException;
 import com.google.common.io.Files;
 import org.apache.commons.lang3.SystemUtils;
+import org.openqa.selenium.firefox.FirefoxBinary;
 import org.openqa.selenium.firefox.FirefoxDriver;
 import org.openqa.selenium.firefox.FirefoxOptions;
 import org.openqa.selenium.firefox.FirefoxProfile;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Properties;
 
 import java.io.File;
 
 public class SeleniumConfig {
-
     private static final Logger LOGGER = LoggerFactory.getLogger(SeleniumConfig.class);
 
+    private static String propertiesPath = String.join(File.separator, "src", "functionaltest", "resources");
+    private static String propertiesFile = "functional-test.properties";
+
+    private static String firefoxBZip2FileUrlLinux = "";
+    private static String firefoxBZip2FileNameLinux = "";
+
     private static File tempDownloadDirectory = Files.createTempDir();
+    private static FirefoxDriver driver;
 
-    public static FirefoxDriver getFirefoxDriver() throws OSNotSupportedException {
-        FirefoxDriver driver;
+    public static FirefoxDriver initFirefoxDriver() throws PropertiesNotLoadedException, OSNotSupportedException {
+        if (driver != null) {
+            return driver;
+        }
+
         FirefoxProfile firefoxProfile = new FirefoxProfile();
-
         firefoxProfile.setPreference("browser.download.folderList", 2);
         firefoxProfile.setPreference("browser.download.dir", tempDownloadDirectory.getPath());
         firefoxProfile.setPreference("browser.helperApps.neverAsk.saveToDisk", "application/json");
 
+        FirefoxOptions firefoxOptions = new FirefoxOptions()
+                .setHeadless(true)
+                .setProfile(firefoxProfile);
+
+        boolean successfullyLoadedProperties = loadProperties();
+        if (!successfullyLoadedProperties) {
+            LOGGER.error("Properties was not properly loaded.");
+            throw new PropertiesNotLoadedException();
+        };
+
         if (SystemUtils.IS_OS_LINUX) {
+            String firefoxBZip2FilePath = String.join(
+                    File.separator, tempDownloadDirectory.getPath(), firefoxBZip2FileNameLinux);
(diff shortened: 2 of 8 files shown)


pom.xml
```diff
@@ -106,6 +106,11 @@
       <version>5.4.1</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+            <groupId>org.rauschig</groupId>
+            <artifactId>jarchivelib</artifactId>
```
a-breid-ericsson: Indention does not wit with the rest of the file.

src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
```diff
@@ -1,31 +1,65 @@
...
+            Utils.extractBZip2InDir(firefoxBZip2FilePath, tempDownloadDirectory.getPath());
+            File firefoxBinaryFilePath = new File(
+                    String.join(File.separator, tempDownloadDirectory.getPath(), "firefox", "firefox"));
+            Utils.makeBinFileExecutable(firefoxBinaryFilePath);
+            FirefoxBinary firefoxBinary = new FirefoxBinary(firefoxBinaryFilePath);
+            firefoxOptions.setBinary(firefoxBinary);
+
             System.setProperty("webdriver.gecko.driver", "src/functionaltest/resources/geckodriver");
         } else if (SystemUtils.IS_OS_WINDOWS) {
             System.setProperty("webdriver.gecko.driver", "src/functionaltest/resources/geckodriver.exe");
```
a-breid-ericsson: Small detail but this OS not supported logging could be LOGGER.error(SystemUtils.OS_NAME + " currently not supported.") There is also an OS_VERSION that could be used.

src/functionaltest/java/com/ericsson/ei/config/Utils.java
```diff
@@ -0,0 +1,117 @@
...
+     * Extracts a BZip2 archive to a destination folder.
+     *
+     * @param firefoxTarballFilePath
+     *            string containing the firefox BZip2 filepath.
+     * @param destination
+     *            string containing a destination path.
+     */
+    public static void extractBZip2InDir(final String firefoxBZip2FilePath, String destinationPath) {
+        LOGGER.info("Extracting firefox BZip2 archive...");
+        try (TarArchiveInputStream fileInput = new TarArchiveInputStream(
```
a-breid-ericsson: Generally you do not want 2 try catch inside each other.

You may catch different errors in the same try catch, either by doing something like this
try {
} catch (someError || AnotherError) {
}
I didn´t get that one to work though, maybe I wrote wrong errors :)
Or like this.
try {
} catch (and error) {
} catch (Another erro5r) {
}

a-breid-ericsson: I see you already do the later onbe so the inner try catch is not nessesary.

src/functionaltest/java/com/ericsson/ei/config/Utils.java
```diff
@@ -0,0 +1,117 @@
...
+                        parent.mkdirs();
+                    }
+                    final FileOutputStream fileOutput = new FileOutputStream(curfile);
+                    IOUtils.copy(fileInput, fileOutput);
+                    fileOutput.close();
+                }
+            } catch (IOException e) {
+                LOGGER.error("IOException.\nError: {}", e.getMessage());
+            }
+            fileInput.close();
```
a-breid-ericsson: this fileInput.close() maybe should be in a finally so wven if it get an error it closes the file.

src/functionaltest/java/com/ericsson/ei/frontend/TestSwitchBackend.java
```diff
@@ -73,6 +74,16 @@ public void testSwitchBackend() throws Exception {
         switchBackendPage.removeInstanceNumber(1);
         assertEquals("switchBackendPage.presenceOfInstance returned true when it should have been false",
             false, switchBackendPage.presenceOfInstance(1));
+
+        tearDownMocks();
+    }
+
+    private void tearDownMocks() {
```
a-breid-ericsson: Small detail, tearDownMocks should be under setupMocks.

Is it not possible to run those functiona from a before and after class? That could make it easier to edit the test in future.

pom.xml
e-edling-ericsson: Done

src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
e-edling-ericsson: Done

src/functionaltest/java/com/ericsson/ei/config/Utils.java
e-edling-ericsson: I took this code from another component but i agree

e-edling-ericsson: I took this code from another component and i guess the reason why it's made this way is that the fileInput needs to be closed after exceptions thrown within. Although apperently in java 7 and onwards try-catches with resources will automatically be closed. Therefore the fileInput.close() should not even be needed.  https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/functionaltest/java/com/ericsson/ei/config/Utils.java
e-edling-ericsson: Removing the close completely, see https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/functionaltest/java/com/ericsson/ei/frontend/TestSwitchBackend.java
e-edling-ericsson: Done

src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
```diff
@@ -1,44 +1,77 @@
...
         firefoxProfile.setPreference("browser.download.folderList", 2);
         firefoxProfile.setPreference("browser.download.dir", tempDownloadDirectory.getPath());
         firefoxProfile.setPreference("browser.helperApps.neverAsk.saveToDisk", "application/json");
 
+        FirefoxOptions firefoxOptions = new FirefoxOptions()
+                .setHeadless(true)
+                .setProfile(firefoxProfile);
+
+        boolean successfullyLoadedProperties = loadProperties();
+        if (!successfullyLoadedProperties) {
```
ned29: simplify this ->if (!loadProperties())

src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
```diff
@@ -49,4 +82,25 @@ public static File getTempDownloadDirectory() {
...
+
+        try {
+            firefoxBZip2FileUrlLinux = properties.getProperty("test.selenium.firefox.BZip2File.linux");
+            firefoxBZip2FileNameLinux = properties.getProperty("test.selenium.firefox.BZip2File.linux.name");
+
+            if (StringUtils.isEmpty(firefoxBZip2FileUrlLinux) || StringUtils.isEmpty(firefoxBZip2FileNameLinux)) {
+                LOGGER.error("Failed to load properties");
+                return false;
+            } else {
+                LOGGER.info("Properties have been loaded.");
```
ned29: set logger to debug level

src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
e-edling-ericsson: I think it's simpler to write it your way but not as descriptive when reading the code. When calling functions in a conditional statement i think it's way clearer to declare a variable with a name that explains the returned value even in a simple case as this. If i'm the only one who thinks that, i'll change it of course.  :)

e-edling-ericsson: Yes it could be simplified, but i think that declaring a variable before with the result from the method makes it way more clearer in a conditional statement such as if-statement even in a simple case such as this one. Don't you agree?

src/functionaltest/resources/functional-test.properties
```diff
@@ -0,0 +1,2 @@
+test.selenium.firefox.BZip2File.linux = https://download.mozilla.org/?product=firefox-61.0.1-ssl&os=linux64&lang=en-GB
```
tobiasake: This is more proper url to use which has no arguments that can cause address lookup issues:
https://ftp.mozilla.org/pub/firefox/releases/61.0.1/linux-x86_64/en-US/firefox-61.0.1.tar.bz2
With this you can extract versions and fileName from the address.
Remove the "......linux.name" and use only url variable.
And add url to the property settings parameter, e.g: "test.selenium.firefox.BZip2File.url.linux "

src/functionaltest/resources/functional-test.properties
```diff
@@ -0,0 +1,2 @@
...
+test.selenium.firefox.BZip2File.linux.name = firefox-61.0.1.tar.bz2
```
tobiasake: This property can be removed and extract filename from the url.

src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
```diff
@@ -49,4 +82,25 @@ public static File getTempDownloadDirectory() {
...
+    private static boolean loadProperties() {
+        final String propertiesFilePath = String.join(File.separator, propertiesPath, propertiesFile);
+        final Properties properties = Utils.getProperties(propertiesFilePath);
+
+        try {
+            firefoxBZip2FileUrlLinux = properties.getProperty("test.selenium.firefox.BZip2File.linux");
+            firefoxBZip2FileNameLinux = properties.getProperty("test.selenium.firefox.BZip2File.linux.name");
+
+            if (StringUtils.isEmpty(firefoxBZip2FileUrlLinux) || StringUtils.isEmpty(firefoxBZip2FileNameLinux)) {
+                LOGGER.error("Failed to load properties");
```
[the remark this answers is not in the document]
e-edling-ericsson: But you don't know which one of them that is not set? I will remove the filename from config so it won't be a problem now though.

src/functionaltest/resources/functional-test.properties
e-edling-ericsson: Done

src/functionaltest/resources/functional-test.properties
e-edling-ericsson: Done

src/functionaltest/java/com/ericsson/ei/config/SeleniumConfig.java
```diff
@@ -83,24 +75,32 @@ public static String getBaseUrl(int randomServerPort) {
         return "http://localhost:" + randomServerPort;
     }
 
+    private static FirefoxBinary installFirefoxBinary() {
+        String[] firefoxBZip2FileUrlLinuxBits = firefoxBZip2FileUrlLinux.split("/");
+        String firefoxBZip2FileNameLinux = firefoxBZip2FileUrlLinuxBits[firefoxBZip2FileUrlLinuxBits.length-1];
+
```
tobiasake: I think you can use FilenameUtils.getName() function for getting the filename from the Url.
Check this page:
https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FilenameUtils.html

e-edling-ericsson: Oh thats seems better, i'll use that thanks :)

PR #51, opened 2018-07-17
Bug ei frontend subscription

sericcson:
Some changes in the code....
But It would be tested with travis before actually merging.


src/functionaltest/java/com/ericsson/ei/frontend/pageobjects/SubscriptionPage.java
```diff
@@ -192,4 +191,8 @@ public void clickReloadLDAP(String response, String responseAuth) throws IOExcep
                 .until(ExpectedConditions.elementToBeClickable(By.id("reloadButton")));
         reloadBtn.click();
     }
+
+    public Boolean textExistsInTable(String loc) {
```
ned29: change this to boolean, unneeded boxing here

src/main/resources/application.properties
```diff
@@ -4,7 +4,7 @@ spring.jackson.default-property-inclusion=non_null
 spring.application.name=ei-frontend
   #${project.artifactId}
 
-server.port=8080
+server.port=8086
```
ned29: move it back to 8080

src/main/resources/application.properties
```diff
@@ -13,7 +13,7 @@ server.port=8080
 
 ######## EI Backend
 ei.frontendServiceHost=localhost
-ei.frontendServicePort=8080
+ei.frontendServicePort=8090
```
ned29: the same

src/main/resources/application.properties
sericcson: Well, I will change it before merging.... right now I am testing(repeatedly running) the code in Travis to find out any failures and hence the causes.

src/functionaltest/java/com/ericsson/ei/frontend/SubscriptionHandlingFunctionality.java
```diff
@@ -56,10 +56,8 @@ public void testSubscription() throws Exception {
         String editButtonXPath = "//tr[td[contains(.,'Subscription1')]]/td/button[contains(text(),'Edit')]";
         String viewButtonXPath = "//tr[td[contains(.,'Subscription1')]]/td/button[contains(text(),'View')]";
         subscriptionPage.clickReload(response);
-        assert (new WebDriverWait(driver, 10)
-                .until((webdriver) -> ((driver.getPageSource().contains("Subscription1")))));
-        assert (new WebDriverWait(driver, 10)
-                .until((webdriver) -> ((driver.getPageSource().contains("Subscription2")))));
+        assert (subscriptionPage.textExistsInTable("//tr[td[contains(.,'Subscription1')]]"));
```
e-edling-ericsson: When using a "generic" interaction with the driver it removes the reason why we chose to go for pageobjects. If there is a change in the html code this would have to be changed in multiple places instead of just 1 place.

What you could do instead of creating basically the same method is:

subscriptionInTable("Subscription1");

And then in subscriptionInTable you just add the strings together so that its possible to choose which subscription but the xpath would be hidden atleast.

e-edling-ericsson: When using a "generic" interaction with the driver it removes the reason why we chose to go for pageobjects. If there is a change in the html code this would have to be changed in multiple places instead of just 1 place.

What you could do instead of creating basically the same method is:

subscriptionInTable("Subscription1");

And then in subscriptionInTable you just add the strings together so that its possible to choose which subscription but the xpath would be hidden atleast.

src/functionaltest/java/com/ericsson/ei/frontend/pageobjects/SubscriptionPage.java
```diff
@@ -192,4 +191,8 @@ public void clickReloadLDAP(String response, String responseAuth) throws IOExcep
                 .until(ExpectedConditions.elementToBeClickable(By.id("reloadButton")));
         reloadBtn.click();
     }
+
+    public boolean textExistsInTable(String loc) {
```
ned29: Do we really need this method? I think you can use buttonExist method where you invoking textExistsInTable

PR #53, opened 2018-07-24
ADD: errors store

ned29:


src/main/resources/static/js/login.js
```diff
@@ -34,6 +34,7 @@ var frontendServiceUrl = $('#frontendServiceUrl').text();
 			},
 			error : function (request, textStatus, errorThrown) {
 				$.jGrowl("Bad credentials", { sticky : false, theme : 'Error' });
+				errorsStore.add("Bad credentials");
```
tobiasake: jGrowl and errorStore can be moved out an function e.g. logMessage(message) , and call this function where we do jGrowl.

ned29: Do you mean create a method with jGrowl and errorsStore where we set message?
Something like
logMessage(message){
jGrowl(message);
errorsStore(message);
}

tobiasake: yes

tobiasake:
Now when you store the log/error message in WebServer java code, I guess all Users WebBrowser sessions see each others error messages in drop down menu list. It should be implemented per WebBrowser sessions and user, so user/web-browser sessions only see the users own error message.

src/main/resources/static/js/errorMessages.js
```diff
@@ -0,0 +1,16 @@
+var frontendServiceUrl = $('#frontendServiceUrl').text();
+const errorsStore = [];
+function viewModel(data){
+    this.errorMessages = ko.observableArray(data);
+}
+function getErrors(){
+    const messageErr = JSON.parse(localStorage.getItem('errorsStore'));
+    var observableObject = $("#alerts")[0];
+    ko.cleanNode(observableObject);
+    ko.applyBindings(new viewModel(messageErr), observableObject);
```
e-edling-ericsson: If there are no errormessages in the localstorage i think there will be problems here(nulls). Maybe that should be handled?

src/main/resources/static/js/errorMessages.js
```diff
@@ -0,0 +1,16 @@
...
+}
+function logMessages(messageErr){
+    $.jGrowl(messageErr, {sticky: false, theme: 'Error'});
+    errorsStore.push({message:messageErr});
+    localStorage.setItem('errorsStore', JSON.stringify(errorsStore));
```
e-edling-ericsson: Don't errorsStore need to retreive the current message list from the localstorage and then append to it? Is the variable "errorstore" keeping the state otherwise?

tobiasake:
Send it out on team mail address so other users can test in different environment and web browser and return feedback if its working or not working before merge.

e-edling-ericsson:
I’ve tried out the implementation aswell and I got a few notes.

•When I open the error notification button where I can see all the error messages. I am not able to read the entire message if they are long
•They don’t show up at all unless I reload the entire webpage.
•When there are no messages, there should maybe not be a small empty div showing up. Maybe it should say something like, “No error messages to be shown” or something better.
•If I click one of the 4 messages I have it just closes down. If i reopen, then I got extremely many messages of that type I clicked on. If I reload the page I got back the 4 messages I had in the beginning. And then I can do the same thing again.
•If the window is full of messages when opening and I click to get a new error message, that one isn’t shown at all.

sericcson:
I have noticed that error dropdown list is not scrollable... you may add scrollbar

sericcson:
It looks good

PR #54, opened 2018-08-13
Test Rule Page Status Implementatuon

sericcson:


src/main/resources/static/js/testrules.js
```diff
@@ -364,6 +364,40 @@ jQuery(document).ready(
...
+    			isEnabled = JSON.parse(ko.toJSON(data)).status;
+    			if(isEnabled == false) {
+    				displayOverlay("Test Rule service is not enabled! To enable it set the backend property [testaggregated.enabled] as [true]")}
+    			},
+    		complete: function () { }
+    	});
+    	// Finish to check backend Test Rule Service status
+    	
+    	function displayOverlay(text) {
+    	    $("<table id='overlay'><tbody><tr><td>" + text + "</td></tr></tbody></table>").css({
```
e-edling-ericsson: Is there no other place where we can put the CSS code? If not i guess this would be fine however I really think we should have files dedicated to CSS so that its easier to change the looks.

e-edling-ericsson: Is there no other place where we can put the CSS code? If not i guess this would be fine however I really think we should have files dedicated to CSS so that its easier to change the looks.

src/main/resources/static/js/testrules.js
```diff
@@ -364,6 +364,40 @@ jQuery(document).ready(
...
+    	        "width": "100%",
+    	        "height": "100%",
+    	        "background-color": "rgba(0,0,0,.5)",
+    	        "z-index": 10000,
+    	        "vertical-align": "middle",
+    	        "text-align": "center",
+    	        "color":  "rgba(128,0,0,1)",
+    	        "font-size": "25px",
+    	        "font-weight": "bold"
+    	    }).appendTo("#testRule");
```
e-edling-ericsson: Is there no other place where we can put the CSS code? If not i guess this would be fine however I really think we should have files dedicated to CSS so that its easier to change the looks.

sericcson: It is done

PR #59, opened 2018-09-10
EI Frontend: Remove Gui functionality that will not be used

sericcson:


src/main/resources/application.properties
```diff
@@ -4,7 +4,7 @@ spring.jackson.default-property-inclusion=non_null
 spring.application.name=ei-frontend
   #${project.artifactId}
 
-server.port=8080
+server.port=8099
```
ned29: move it back

src/main/resources/application.properties
```diff
@@ -13,7 +13,7 @@ server.port=8080
 
 ######## EI Backend
 ei.frontendServiceHost=localhost
-ei.frontendServicePort=8080
+ei.frontendServicePort=8099
```
ned29: the same

.gitignore
```diff
@@ -16,6 +16,10 @@ log/
 target/
 
 # Others
+<<<<<<< HEAD
```
ned29: i think this is, should be removed and below also(seems like unresolved merge conflict)

PR #61, opened 2018-09-11
Remove Gui functionality that will not be used

saif-ericsson:


sericcson:
Tobias: Your first point is valid regarding my story and I fix it. 
The other two points are regarding Andrey's implementation. As his story is closed and merged, I can create a bug story for that?

eroohen:
Just a minor point: Never refer to Ericsson-internal JIRA stories in github. You don' have to worry about it now as I already removed references to these in the title.

PR #62, opened 2018-09-12
Remove bugged tooltip and redraw table on sidenav collapse

Christoffer-Cortes:

### DIFF
--- a/src/main/resources/static/js/main.js
+++ b/src/main/resources/static/js/main.js
@@ -17,36 +17,43 @@ jQuery(document).ready(function() {
 	var frontendServiceUrl = $('#frontendServiceUrl').text();
 
 	function loadMainPage() {
+		$("#navbarResponsive").removeClass("show");
 	    $("#selectInstances").visible();
 		$("#mainFrame").load("subscriptionpage.html");
 	}
 
 	$("#testRulesBtn").click(function() {
+		$("#navbarResponsive").removeClass("show");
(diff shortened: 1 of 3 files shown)


src/main/resources/static/js/main.js
```diff
@@ -17,36 +17,43 @@ jQuery(document).ready(function() {
 	var frontendServiceUrl = $('#frontendServiceUrl').text();
 
 	function loadMainPage() {
+		$("#navbarResponsive").removeClass("show");
```
a-breid-ericsson: Why is the indention here ...
```
Line one
        Line 2
    Line 3
        Line4
```
?

Christoffer-Cortes: It's caused by the use of tab in the source code... it hasn't been formatted correctly.

src/main/resources/static/js/subscription.js
```diff
@@ -414,6 +414,10 @@ jQuery(document).ready(function () {
             }
         }
     });
+
+    $("#sidenavCollapse").click(function() {
+		table.ajax.reload();
```
tobiasake: Does this mean that Datatable is reloaded and it re-retrieve data from backend, e.g. it makes a Get all subscriptions request everytime sidebar is collapsed?

Christoffer-Cortes: Apparently the function was unable to detect dataset changes but I've now changed it with a better method.

PR #69, opened 2018-10-16
Subscription GUI: added AND  & ARE between requirements and conditions

saif-ericsson:


src/functionaltest/java/com/ericsson/ei/frontend/SubscriptionHandlingFunctionality.java
```diff
@@ -204,12 +204,24 @@ public void testSubscription() throws Exception {
         String requirementFieldID = "requirementID";
 
         subscriptionPage.selectDropdown(selectAuthID, authValue);
-
         subscriptionPage.addFieldValue(userNameID, userName);
         subscriptionPage.addFieldValue(tokenID, token);
         String kvID = "kvID";
-        subscriptionPage.clickKVbtn(kvID);
+        subscriptionPage.clickKVbtn(kvID);
```
a-breid-ericsson: Trailing white spaces here.
It is possible to automatically remove trailing white spaces in eclipse when you save.

src/main/resources/templates/subscription.html
```diff
@@ -240,16 +240,24 @@ <h3 class="modal-title text-center" id="formHeader">Subscription Form</h3>
...
                             <!-- ko foreach:   { data: requirements, as: 'requirements_item' } -->
                             <div class="pl-0 ml-0">
+                                <!-- ko if: $index() !== 0 -->
+                                <h5 align="center">OR</h5>
+                                <!-- /ko -->
                                 <img data-toggle="tooltip" data-placement="top" width="16" alt="Information" src="assets/images/information.png" title="Info: 'AND' is used between Conditions in Requirement groups, and 'OR' is used between Requirement groups"
                                 />
                             </div>
                             <div class="p-1 border border-primary form-group">
+
```
a-breid-ericsson: spaces.

src/main/resources/templates/subscription.html
```diff
@@ -240,16 +240,24 @@ <h3 class="modal-title text-center" id="formHeader">Subscription Form</h3>
...
                             </div>
                             <div class="p-1 border border-primary form-group">
+                            
                                 <label class="control-label font-weight-bold">Requirement</label>
                                 <div id="requirementID">
 
                                     <!-- ko foreach: requirements_item.conditions -->
                                     <div id="conditionID">
-                                        <label class="control-label font-weight-bold">Condition</label>
+										<!-- ko if: $index() !== 0 -->
```
a-breid-ericsson: Invalid indention here.

src/main/resources/templates/subscription.html
```diff
@@ -240,16 +240,24 @@ <h3 class="modal-title text-center" id="formHeader">Subscription Form</h3>
...
+										<h6 align="center">AND</h6>
```
a-breid-ericsson: Adding AND when editing subscriptions looks strange.

a-breid-ericsson: I would recommend to remove AND and instead add $index + 1 to Cindition.
So it will be:
Condition 1:
  XXXXXXX
Condition 2:
  ZZZZZZZZ

evasiba-ericsson: The purpose of the story is to add make clear the logical operations performed, that all conditions in a requirement needs to be fulfilled for a subscription to trigger.

a-breid-ericsson: This may look cleare from a coding perspective, but are all that use EI coders?

Then the "Condition" text should be removed for every condition and instead display as a list of conditions.

PR #71, opened 2018-10-18
Improved error handeling and added some code comments.

a-breid-ericsson:
**Fixes issues when adding subscription in EI front end.**

- fixed bug name validation failed, name was only validated based on the last character, thus the name could be "My in%vali¤Name" and still pass.
- Improved error responses to user when entering invalid or no information.
- Added some code comments.


src/main/resources/static/js/subscription.js
```diff
@@ -739,85 +739,95 @@ jQuery(document).ready(function () {
...
-            }
-        */
-
-        if (vm.subscription()[0].subscriptionName() == "") {
-            window.logMessages("Error: SubscriptionName field must have a value");
-            return;
-        }
+        // Validate notificationType field
         if (vm.subscription()[0].notificationType() == null) {
             window.logMessages("Error: notificationType field must boolean a value");
```
e-pettersson-ericsson: This log does not make sense. Maybe have "Error: notificationType must have a value"?

src/main/resources/templates/subscription.html
```diff
@@ -82,7 +82,7 @@ <h3 id="subData">Subscription Data</h3>
                         <h3 class="modal-title text-center" id="formHeader">Subscription Form</h3>
                     </div>
                 </div>
-                <div class="modal-body form" align="left">
+                <div class="modal-body form" allign="left">
```
e-pettersson-ericsson: Misspelled? Should be "align"

src/main/resources/templates/subscription.html
```diff
@@ -234,12 +256,14 @@ <h3 class="modal-title text-center" id="formHeader">Subscription Form</h3>
                                                optionsText: $data.repeat(),
                                                value: $data.repeat,
                                                optionsCaption: 'Choose...'"></select>
+                                    <font class="hiddenField" id="repeatNotSet" color="red" size="2">> Repeat must be set</font>
```
e-pettersson-ericsson: One too many '>'

src/main/resources/templates/subscription.html
a-breid-ericsson: Should be, but was marked red in Vs code, and OP with allign, think it does not matter so maybe to remove it.

src/main/resources/templates/subscription.html
a-breid-ericsson: No the > is part of the message :)
Will looke like this [Dropdown](in re dcolor)> Repeat must be set

src/main/resources/static/js/subscription.js
a-breid-ericsson: Maybe to remove those messages since it error now appear in at the field where the values was added.

e-edling-ericsson:
I think it would be great to know what kind of errors you have fixed in the PR. I think i could give a better review with that information :) I could with that information download the code and try those things out and see how it looks like in the browser.

src/main/resources/static/js/subscription.js
e-pettersson-ericsson: I think it's fine to keep both (as long as the information is the same) to clearly highlight that something is wrong.

src/main/resources/static/css/style.css
```diff
@@ -55,6 +55,9 @@
     height: 45px;
     animation: spin 1s cubic-bezier(.86,0,.07,1) infinite;
 }
+.hiddenField {
```
Christoffer-Cortes: There is a built in class in Bootstrap called 'd-none' that could be used.

PR #72, opened 2018-10-22
Removal of Information button 

saif-ericsson:


src/main/resources/templates/index.html
```diff
@@ -54,12 +54,6 @@
             <span class="nav-link-text">Test Rules</span>
           </a>
         </li>
-        <li id="eiInfoBtn" class="nav-item" data-toggle="tooltip" data-placement="right" title="Eiffel Intelligence Information">
```
a-breid-ericsson: This is not the button I mentioned in the bug story. I will update the story.

saif-ericsson: Well this button is not showing content.

The button you mentioned is information icon.... a common way for user information. It inform user about the use of AND and OR condition. I think it should be there for user help.  Do you still think that it should be removed...?

saif-ericsson:
Closing the pull request as, after discussion, it is agreed that this functionality need not to be implemented.

PR #89, opened 2018-11-05
fixed: upload rules and upload events do not need manual refreshing for any browser

saif-ericsson:
### Applicable Issues
In Eiffel-Intelligence Frontend, on test rule page, rules upload and event upload is not working with any browser except Firefox until page is manually refreshed every time a rule/rules event/events need to be uploaded.

### Description of the Change
It was figured out that Java Script method onchage() was not detecting any change when the same file (containing rules or events) is uploaded again. In fact the name of the file need to be different to get  onchange() method triggered. So a code line was added which remove the file name/path everytime a change get triggered.

### Benefits
Now upload rules and upload events are compatible with all browsers.

### Possible Drawbacks
NO


src/main/resources/templates/subscription.html
```diff
@@ -270,16 +270,20 @@ <h3 class="modal-title text-center" id="formHeader">Subscription Form</h3>
                                 <!-- ko if: $index() !== 0 -->
                                 <h5>OR</h5>
                                 <!-- /ko -->
+<<<<<<< HEAD
```
Christoffer-Cortes: This part here should be removed. Remains from a merge conflict.

PR #90, opened 2018-11-06
 Fix EI front end uploading file problem in Chrome

J-Binieda-ericsson:
### Applicable Issues
When editing a rule in test rules page, and then uploading more rules from a json-file all rules become duplicates of the first edited rule.

### Description of the Change
Rule biniding list and event list updates now and works properly in chrome. Previously all contet was removed, replaced and list was created with new content instead of being updated.

I merged two similar functions into one  to make code shorter and more readable.

### Alternate Designs
None

### Benefits
It is possible now to upload test rules in Chrome, even if you modified first rule before.

### Possible Drawbacks
None

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: Jakub Binieda, jakub.binieda@ericsson.com

### DIFF
--- a/src/main/resources/static/js/testrules.js
+++ b/src/main/resources/static/js/testrules.js
@@ -82,7 +82,7 @@ jQuery(document).ready(
           self.rulesBindingList.splice(context.$index(), 1);
           if (self.rulesBindingList().length == 0) {
             window.logMessages("Deleted all rule types, but we need atleast one Rule type, Here add default rule type");
-            self.rulesBindingList.push(ruleTemplate);
+            self.rulesBindingList.push(JSON.parse(JSON.stringify(ruleTemplate)));
           }
         };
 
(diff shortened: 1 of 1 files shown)


src/main/resources/static/js/testrules.js
```diff
@@ -221,14 +190,19 @@ jQuery(document).ready(
...
-            vm.eventsBindingList.removeAll();
-            $('.eventsListDisplay > div:gt(0)').remove();
-            vm.eventsBindingList = ko.observableArray(eventsList);
-            ko.applyBindings(vm, $("#testEventsDOMObject")[0]);
-            closeTooltip();
-            loadTooltip();
+
+            var list = JSON.parse(fileContent);
+
+            if (type == true) {
```
Christoffer-Cortes: Perhaps call this boolean something like 'isRules'

src/main/resources/static/js/testrules.js
```diff
@@ -221,14 +190,19 @@ jQuery(document).ready(
...
+                vm.rulesBindingList([]);
+                vm.rulesBindingList(list);
+                vm.rulesBindingList.valueHasMutated();
```
Christoffer-Cortes: I'm not sure if valueHasMutated is even needed. It's used in subscription but try removing it and see if it affects anything.

src/main/resources/static/js/testrules.js
J-Binieda-ericsson: Done

src/main/resources/static/js/testrules.js
J-Binieda-ericsson: It is used to force the update of an observable, but it seems that it is not needed here. I removed it.

PR #99, opened 2018-11-16
Change behaviour of the EI backendInstancesinformation file

e-edling-ericsson:
### Description of the Change

EIBackendInstanceInformation.json was not stored outside the application with resulted in issues when running as a war file since we can't write to one. EIBackendInstancesInformation.json file will now be stored in .eiffel folder which will automatically be created in USER home. If user home does not exist the user may change the default behaviour in application.json.
### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: 

### DIFF
--- a/src/main/java/com/ericsson/ei/frontend/utils/BackEndInstanceFileUtils.java
+++ b/src/main/java/com/ericsson/ei/frontend/utils/BackEndInstanceFileUtils.java
@@ -94,6 +94,7 @@ public class BackEndInstanceFileUtils {
 
             if (!fileContainsJsonArray()) {
                 LOG.error("File does not contain valid json! JSON:'" + new String(Files.readAllBytes(Paths.get(eiInstancesPath))) + "'.");
+                System.exit(-1);
             }
         } catch(Exception e) {
             String message = String.format(
--- a/src/main/java/com/ericsson/ei/frontend/utils/BackEndInstanceFileUtils.java
+++ b/src/main/java/com/ericsson/ei/frontend/utils/BackEndInstanceFileUtils.java
@@ -24,38 +24,31 @@ public class BackEndInstanceFileUtils {
     private static final Logger LOG = LoggerFactory.getLogger(BackEndInstanceFileUtils.class);
 
     private static final String BACKEND_INSTANCES_DEFAULT_FILENAME = "EIBackendInstancesInformation.json";
-    private static final String EI_HOME_DEFAULT_NAME = ".eiffel";
+    private static final String EI_HOME_DEFAULT_NAME = ".eiffel-intelligence-frontend";
 
     private String eiInstancesPath;
-
-    @Value("${ei.backendInstancesFileName:#{null}}")
-    private String eiInstancesFileName;
-
-    @Value("${ei.home}")
     private String eiHome;
+
+    @Value("${ei.backendInstancesFilePath:#{null}}")
+    private String backendInstancesFilePath;
 
     @PostConstruct
     public void init() throws IOException {
         LOG.info("Initiating BackEndInstanceFileUtils.");
 
-        // If user did not choose ei home folder, set EI home folder to default
-        if(eiHome == null || eiHome.isEmpty()) {
+        // Use home folder if a specific backendInstancesFilePath isn't provided
+        if(backendInstancesFilePath == null || backendInstancesFilePath.isEmpty()) {
             String homeFolder = System.getProperty("user.home");
             eiHome = Paths.get(homeFolder, EI_HOME_DEFAULT_NAME).toString();
-        }
+            Boolean eiHomeExists = Files.isDirectory(Paths.get(eiHome));
+            if (!eiHomeExists) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/ericsson/ei/frontend/utils/BackEndInstanceFileUtils.java
```diff
@@ -22,16 +22,34 @@
 public class BackEndInstanceFileUtils {
 
     private static final Logger LOG = LoggerFactory.getLogger(BackEndInstanceFileUtils.class);
-    private static final String PATH_TO_WRITE = "src/main/resources/EIBackendInstancesInformation.json";
+
+    private static final String BACKEND_INSTANCES_FILENAME = "EIBackendInstancesInformation.json";
+    private static final String EI_HOME_DEFAULT_NAME = ".eiffel";
 
     @Value("${ei.backendInstancesPath:#{null}}")
```
a-breid-ericsson: As long as the value ei.backendInstancesPath is set in application.properties this will write to src/main/resources/EIBackendInstancesInformation.json 

Maybe you should remove the @Value("${ei.backendInstancesPath:#{null}}") and then the line in application.properties. Since it is replaced with ei.home.

src/main/java/com/ericsson/ei/frontend/utils/BackEndInstanceFileUtils.java
```diff
@@ -92,4 +110,13 @@ private boolean fileContainsJsonArray() {
         }
     }
 
+    private void createEiHomeFolder() {
+        Boolean success = (new File(eiHome)).mkdirs();
+
+        if (!success) {
+           LOG.error("Failed to create eiffel intelligence home folder in {}. Please check access rights or change default folder in application.properties.".format(eiHome));
+           System.exit(-1);
```
a-breid-ericsson: Maybe this should raise an Exception?

src/main/java/com/ericsson/ei/frontend/utils/BackEndInstanceFileUtils.java
e-edling-ericsson: backendInstancesPath does not exist anymore, its just called "backendInstancesFileName" and it's just there so that you can choose filename if you would have multiple files in the eiffell home folder for some reason.

ei.home just tells the system where the user configuration folder is placed.

No it won't write to src/main/resources/EIBackendInstancesInformation.json. It will write in the EI home path which by default is user.home/.eiffel.

src/main/java/com/ericsson/ei/frontend/utils/BackEndInstanceFileUtils.java
e-edling-ericsson: Ye that might be better, i have no idéa which way to handle this the best, it should atleast not start the application if it's not possible to load/create user configurations.

src/main/java/com/ericsson/ei/frontend/utils/BackEndInstanceFileUtils.java
a-breid-ericsson: I see that you changed the code since my comment.

src/main/java/com/ericsson/ei/frontend/utils/BackEndInstanceFileUtils.java
```diff
@@ -70,16 +86,34 @@ public void dumpJsonArray(JsonArray jsonArrayToDump) {
...
-            Files.write(Paths.get(eiInstancesPath), "[]".getBytes());
-            return;
+        try {
+            if (!(new File(eiInstancesPath).isFile())) {
+                createFileWithDirs();
+                return;
+            }
+
+            if (!fileContainsJsonArray()) {
+                LOG.error("File does not contain valid json! JSON:'" + new String(Files.readAllBytes(Paths.get(eiInstancesPath))) + "'.");
```
a-breid-ericsson: If the File does not contain valid JSON then maybe we should also do a System.exit.
Note that if file is empty "" we could add just [] and it would be valid. This is if user creates an empty file or manually deletes all content.

src/main/java/com/ericsson/ei/frontend/utils/BackEndInstanceFileUtils.java
```diff
@@ -70,16 +86,34 @@ public void dumpJsonArray(JsonArray jsonArrayToDump) {
...
+            String message = String.format(
+                    "Failed to read backendInstancesFilePath %s. Please check access rights or choose another backendInstancesFilePath in application.properties.", eiInstancesPath);
+            LOG.error(message);
+            System.exit(-1);
         }
+    }
 
-        if (!fileContainsJsonArray()) {
-            LOG.error("File does not contain valid json! JSON:'" + new String(Files.readAllBytes(Paths.get(eiInstancesPath))) + "'.");
+    private void createFileWithDirs() throws IOException {
```
a-breid-ericsson: Looks much better :)

PR #104, opened 2018-11-21
Add routing capability

Christoffer-Cortes:
### Applicable Issues
When refreshing a page you always get directed to the subscription page and there is no way to directly reach a specific page.

### Description of the Change
- Navigo lib has been added
- Routing paths have been added
- Login routing will redirect if security is false or if a user is already logged in
- Security and login is now checked in main.js
- References to jmesPathRulesSetUp have been removed

### Alternate Designs
One of the various routing libs that exist on github.

### Benefits
This adds routing capability so that a user can refresh and stay on the same page or open a specific page directly.

### Possible Drawbacks
--

### Sign-off

    indicated in the file; or

    in the file; or

    it.


src/main/resources/static/assets/navigo/7.1.2/navigo.min.js
```diff
@@ -0,0 +1,2 @@
...oid 0===n?"undefined":e(n))?{route:t,handler:n.uses,name:n.as,hooks:o||n.hooks}:{route:t,handler:n,hooks:o}),this._add},_getRoot:function(){return null!==this.root?this.root:(this.root=a(this._cLoc().split("?")[0],this._routes),this.root)},_listen:function(){var e=this;if(this._usePushState)window.addEventListener("popstate",this._onLocationChange);else if("undefined"!=typeof window&&"onhashchange"in window)window.addEventListener("hashchange",this._onLocationChange);else{var t=this._cLoc(),n=void 0,o=void 0;(o=function(){n=e._cLoc(),t!==n&&(t=n,e.resolve()),e._listeningInterval=setTimeout(o,200)})()}},_cLoc:function(){return"undefined"!=typeof window?void 0!==window.__NAVIGO_WINDOW_LOCATION_MOCK__?window.__NAVIGO_WINDOW_LOCATION_MOCK__:o(window.location.href):""},_findLinks:function(){return[].slice.call(document.querySelectorAll("[data-navigo]"))},_onLocationChange:function(){this.resolve()},_callLeave:function(){var e=this._lastRouteResolved;e&&e.hooks&&e.hooks.leave&&e.hooks.leave(e.params)}},n.PARAMETER_REGEXP=/([:*])(\w+)/g,n.WILDCARD_REGEXP=/\*/g,n.REPLACE_VARIABLE_REGEXP="([^/]+)",n.REPLACE_WILDCARD="(?:.*)",n.FOLLOWED_BY_SLASH_REGEXP="(?:/$|$)",n.MATCH_REGEXP_FLAGS="",n});
```
a-breid-ericsson: Do we need to foss this file?

src/main/resources/static/js/main.js
```diff
@@ -70,18 +70,20 @@ jQuery(document).ready(function() {
     });
 
     function doIfUserLoggedIn(currentUser) {
-        if (currentUser != "") {
```
J-Binieda-ericsson: Do we not need this check?

src/main/resources/static/js/add-instances.js
```diff
@@ -2,6 +2,8 @@ jQuery(document).ready(function() {
 var frontendServiceUrl = $('#frontendServiceUrl').text();
 var frontendServicePath = "/backend";
 function instanceModel() {
+    var router = new Navigo(null, true, '#');
```
tobiasake: Can you add a comment line here about what Navigo do?

src/main/resources/static/js/main.js
Christoffer-Cortes: You never enter this function unless a successful login request was made and that request can only be successful if a user has once logged in with the right credentials. This means that it must have a value if it was successful so I didn't feel it was needed to check whether it was empty or not.

src/main/resources/static/js/add-instances.js
Christoffer-Cortes: The router is initialized in a few places though, add-instances, switch-instances, login, and main. Should I comment in each file or perhaps only in main where the routes are defined?

Christoffer-Cortes: I chose to put the comment in main.js where the routes are first defined.

tobiasake: Ok, I think we should have some comment about it, so users that have not worked with the code earlier can easier understand why Navigo is used.

PR #125, opened 2018-12-07
Versions fix

a-breid-ericsson:
Application information such as application name and version is stored in the default-application.properties instead of the application.properties that may be over written by users. This enables us to make use of those values in the front end and display the correct versions.

There is also new fields in the application.properties, build.version.name and =build.version is seperated from the version displayed from the pom, the users may add their own version name if verison is set, if version is set without name then it will be displayed as Enterprise Version in EI front end.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:


src/main/resources/static/js/eiInfo.js
```diff
@@ -1,9 +1,18 @@
 jQuery(document).ready(function() {
     // Fetch injected URL from DOM
-    var frontendServiceUrl = $('#frontendServiceUrl').text();
-    var backendServerUrl = $('#backendServerUrl').text();
+    var frontEndServiceUrl = $('#frontendServiceUrl').text();
+    var frontEndVersion = $('#version').text();
+    var frontEndEnterpriseVersion = $('#enterpriseVersion').text();
+    var frontEndEnterpriseVersionName = $('#enterpriseVersionName').text();
```
e-edling-ericsson: What do you mean with Enterprise version? This seems mixed with the concept of community/enterprise version of softwares. Maybe this should change name to not cause confusion?

src/main/java/com/ericsson/ei/frontend/utils/WebControllerUtils.java
```diff
@@ -36,30 +40,42 @@
...
 
+    private String applicationName;
+
     @Autowired
     private BackEndInstancesUtils backEndInstancesUtils;
 
+    @PostConstruct
+    public void init() throws IOException {
+        Properties properties = new Properties();
+        properties.load(WebControllerUtils.class.getResourceAsStream("/default-application.properties"));
```
tobiasake: have you test executed it with the war file ? : java -jar ei.war , so the classpath works in all situations.

src/main/java/com/ericsson/ei/frontend/WebController.java
```diff
@@ -54,7 +54,9 @@ public String testRules(Model model) {
     @RequestMapping("/eiInfo.html")
     public String eiInfo(Model model, HttpServletRequest request) {
         model.addAttribute("frontendServiceUrl", frontEndUtils.getFrontEndServiceUrl());
-        model.addAttribute("frontendServiceVersion", frontEndUtils.getVersion());
+        model.addAttribute("version", frontEndUtils.getVersion());
+        model.addAttribute("enterpriseVersion", frontEndUtils.getEnterpriseVersion());
+        model.addAttribute("enterpriseVersionName", frontEndUtils.getEnterpriseVersionName());
```
tobiasake: Why do we need version name?

src/main/java/com/ericsson/ei/frontend/utils/WebControllerUtils.java
a-breid-ericsson: Yes.

src/main/java/com/ericsson/ei/frontend/WebController.java
a-breid-ericsson: It is added as an optional parameter in the application.properties so the sys-admins may enter an internal version, what ever they want to call that version. if that version is left empty it will not be displayed. You where here when we discussed it.

a-breid-ericsson: Parameter removed.

src/main/resources/static/js/eiInfo.js
a-breid-ericsson: Done.

PR #127, opened 2018-12-07
Fix log issue when adding backend instance

e-edling-ericsson:
### Applicable Issues

The frontend webserver did not return correct response when adding backend instances.
### Description of the Change

It should not return the correct message structure which means that the front end now can print out the error message.
### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: 

### DIFF
--- a/src/main/java/com/ericsson/ei/frontend/utils/BackEndInformationControllerUtils.java
+++ b/src/main/java/com/ericsson/ei/frontend/utils/BackEndInformationControllerUtils.java
@@ -22,6 +22,7 @@ import java.util.stream.Collectors;
 
 import javax.servlet.http.HttpServletRequest;
 
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -37,9 +38,9 @@ import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 
 @Component
-public class BackEndInfoirmationControllerUtils {
+public class BackEndInformationControllerUtils {
 
-    private static final Logger LOG = LoggerFactory.getLogger(BackEndInfoirmationControllerUtils.class);
+    private static final Logger LOG = LoggerFactory.getLogger(BackEndInformationControllerUtils.class);
 
     @Autowired
     private BackEndInstancesUtils backEndInstancesUtils;
@@ -156,8 +157,9 @@ public class BackEndInfoirmationControllerUtils {
                         getHeaders(), HttpStatus.BAD_REQUEST);
             }
         } catch (Exception e) {
+            LOG.error(ExceptionUtils.getStackTrace(e));
             return new ResponseEntity<>(
-                    "Internal error, " + e.getMessage(),
+                    "{\"message\": \"Internal server error.\"}",
                     getHeaders(), HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/ericsson/ei/frontend/utils/BackEndInformationControllerUtils.java
```diff
@@ -156,8 +157,9 @@
                         getHeaders(), HttpStatus.BAD_REQUEST);
             }
         } catch (Exception e) {
+            LOG.error(ExceptionUtils.getStackTrace(e));
```
tobiasake: It would be good to have e.getMessage... also,, so we can get the error text message in the log error message.

src/main/java/com/ericsson/ei/frontend/utils/BackEndInformationControllerUtils.java
```diff
@@ -156,8 +157,9 @@
...
             return new ResponseEntity<>(
-                    "Internal error, " + e.getMessage(),
+                    "{\"message\": \"Internal server error.\"}",
```
tobiasake: Same comment here as well,, give some more info about what the issue is back to the user.

e-edling-ericsson: The thing is, this issue is not relevant to the browser as nothing should have been wrong with the contents from the browser, the other error messages is for that. If it comes to the point where we have an internal server error it should be enough to say that there was an internal server error if the problem is some sort of unhandled issue in the webserver. If it ever comes to the point where this happens, there's a bug that we should fix anyways and then they have the log to report this with.

src/main/java/com/ericsson/ei/frontend/utils/BackEndInformationControllerUtils.java
e-edling-ericsson: the stacktrace contains the message as well so there should be no need for that :)

src/main/java/com/ericsson/ei/frontend/utils/BackEndInformationControllerUtils.java
a-breid-ericsson: I agree with Erik here, the e-messages is for admins and not users. In a big applications all errors would be handled and given an error code that the user could report further.

e-edling-ericsson: When it comes to the internal server error, the user cannot not solve it without access to the deployed system. It is a bug or environmental issues in the back-end as it is an "internal server error". Therefore it's not something that the user have done wrong, it's an error in the server itself. If there however is anything wrong with the request the user sends, the user should get an answer of what was wrong in their request, and this is done in the code. 

This also follows some standard REST API error handling best practices as mentioned in https://blog.restcase.com/rest-api-error-codes-101/
4xx: Client Error - This category of error status codes points the finger at clients.
5xx: Server Error - The server takes responsibility for these error status codes.

Also, sending stacktrace information back to the user in a REST system can also result in exposing data/structure from for example the database or other "secret" information in the deployed system. 

If we send something back to the browser about an internal error i think it should be done as Anders mentioned, send an error code and document the error codes.

evasiba-ericsson: Often an exception has a message with it so use that message in the ResponseEnity if it exist, otherwise use the one you use now.

e-edling-ericsson:
I will remove this PR since it was implemented in another merge recently. And about the error handling, this has been discussed and will be handled in a later PR. Basically so that we have a structured way of sending this back to the user and give relevant error messages depending on what type of issue they are facing. As a product, there are differences you want to have if it's a 400 error or 500 error.

PR #132, opened 2018-12-20
Fix multiple user login problem

a-breid-ericsson:
There was an issue where only one user could login to the EI back end using the front-end. This was caused by the front-end reusing the same clienbt for each call, and when a user did login it was the clients session that became logged in, and all users of a front-end used the same client.

### Applicable Issues
Only one user could be logged in at the same time.
Mockito was not working well in windows.

### Description of the Change
Login is no longer bound the the bridge client, but a new client is createt for each bridged calls. An X-Auth-Token is saved for each users session in the bridge and is reused to authenticate against the back end for as long as the user is logged in.

Tests uses mockserverclient instead if Mockito for handeling http request calls in tests.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:

### DIFF
--- a/src/main/java/com/ericsson/ei/frontend/EIRequestsController.java
+++ b/src/main/java/com/ericsson/ei/frontend/EIRequestsController.java
@@ -52,8 +52,8 @@ import com.ericsson.ei.frontend.utils.EIRequestsControllerUtils;
 public class EIRequestsController {
 
     private static final Logger LOG = LoggerFactory.getLogger(EIRequestsController.class);
-    private static final String X_PATH_NAME = "x-auth-token";
-    List<String> HEADERS_TO_COPY = new ArrayList<>(Arrays.asList(X_PATH_NAME, "authorization"));
+    private static final String X_AUTH_TOKEN = "x-auth-token";
+    List<String> HEADERS_TO_COPY = new ArrayList<>(Arrays.asList(X_AUTH_TOKEN, "authorization"));
 
     @Autowired
     private EIRequestsControllerUtils eiRequestsControllerUtils;
@@ -67,11 +67,11 @@ public class EIRequestsController {
         "/auth/*", "/queryAggregatedObject", "/queryMissedNotifications", "/query", "/rules/rule-check/testRulePageEnabled" }, method = RequestMethod.GET)
     public ResponseEntity<String> getRequests(Model model, HttpServletRequest request) {
         String eiRequestUrl = eiRequestsControllerUtils.getEIRequestURL(request);
-        HttpGet eiRequest = new HttpGet(eiRequestUrl);
+        HttpGet outgoingRequest = new HttpGet(eiRequestUrl);
 
-        eiRequest = (HttpGet) addHeadersToRequest(eiRequest, request);
+        outgoingRequest = (HttpGet) addHeadersToRequest(outgoingRequest, request);
 
-        return getResponse(eiRequest, request);
+        return getResponse(outgoingRequest, request);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/resources/static/js/subscription.js
```diff
@@ -552,11 +551,10 @@ jQuery(document).ready(function () {
                 reload_table();
             },
             error: function (XMLHttpRequest, textStatus, errorThrown) {
-                window.logMessages(XMLHttpRequest.responseText);
                 reload_table();
                 var responseJSON = JSON.parse(XMLHttpRequest.responseText);
                 for (var i = 0; i < responseJSON.length; i++) {
-                    $.jGrowl(responseJSON[i].subscription + " :: " + responseJSON[i].reason, { sticky: true, theme: 'Error' });
+                    window.logMessages("Error deleteing subscription: [" + responseJSON[i].subscription + "] Reson: [" + responseJSON[i].reason + "]");
```
Christoffer-Cortes: 'Reason'

src/main/java/com/ericsson/ei/frontend/EIRequestsController.java
```diff
@@ -48,8 +52,8 @@
 public class EIRequestsController {
 
     private static final Logger LOG = LoggerFactory.getLogger(EIRequestsController.class);
-
-    private CloseableHttpClient client = HttpClientBuilder.create().build();
+    private static final String X_PATH_NAME = "x-auth-token";
```
Christoffer-Cortes: Couldn't this simply be called X_AUTH or X_AUTH_TOKEN?

tobiasake: Is "x-auth-token" HTTP Standard?
I think also that we should have "auth" in the varaible name.

src/main/java/com/ericsson/ei/frontend/EIRequestsController.java
```diff
@@ -64,13 +68,10 @@
...
         HttpGet eiRequest = new HttpGet(eiRequestUrl);
-        String header = request.getHeader("Authorization");
 
-        if (header != null) {
-            eiRequest.addHeader("Authorization", header);
-        }
+        eiRequest = (HttpGet) addHeadersToRequest(eiRequest, request);
 
-        return getResponse(eiRequest);
+        return getResponse(eiRequest, request);
```
tobiasake: Why two different two requests?
Don't add headers to the existing request?

tobiasake: Same comment on the other places in the code where this function is called.

src/main/resources/static/js/subscription.js
```diff
@@ -1090,7 +1088,8 @@ jQuery(document).ready(function () {
                 var errors = "";
                 for (var i = 0; i < responseJSON.length; i++) {
                     errors = errors + "\n" + responseJSON[i].reason;
-                    $.jGrowl(responseJSON[i].subscription + " :: " + responseJSON[i].reason, { sticky: true, theme: 'Error' });
+                    // Errors are displayed in the form.
+                    // $.jGrowl(responseJSON[i].subscription + " :: " + responseJSON[i].reason, { sticky: true, theme: 'Error' });
```
tobiasake: Remove this commented out code if it is not going to be used.

src/main/java/com/ericsson/ei/frontend/EIRequestsController.java
Christoffer-Cortes: No, it's an unregistered header and is missing here: https://www.iana.org/assignments/message-headers/message-headers.xhtml.
It's used in spring for session authentication and every other app can have their own definition of its use.

src/main/resources/static/js/subscription.js
a-breid-ericsson: Fixed.

src/main/java/com/ericsson/ei/frontend/EIRequestsController.java
a-breid-ericsson: Fixed.

src/main/java/com/ericsson/ei/frontend/EIRequestsController.java
a-breid-ericsson: Fixed.

src/main/resources/static/js/subscription.js
a-breid-ericsson: Fixed.

PR #140, opened 2019-01-09
Implement token auth

a-breid-ericsson:
Since for jenkins both username password and token may be needed it may be easier to add it as it´s own field. It could be added as a parameter in the key/value pairs as well. But this clerifies the use of token more.

Before this may work the implementation in EI back-end needs to merged as well.
 

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:

### DIFF
--- a/src/main/resources/templates/subscription.html
+++ b/src/main/resources/templates/subscription.html
@@ -251,7 +251,7 @@
                                     </select>
                                 </div>
                             </div>
-                            <div class="p-1 border form-group" data-bind="visible: authenticationType() == 'BASIC_AUTH'">
+                            <div class="p-1 border form-group" data-bind="visible: authenticationType() == 'BASIC_AUTH' || authenticationType() == 'BASIC_AUTH_JENKINS' ">
                                 <label class="pt-1 control-label font-weight-bold">Username</label>
                                 <div>
                                     <input id="userNameInput" title="Enter user name" data-bind="textInput:userName" name="userName"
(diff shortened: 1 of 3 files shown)


src/main/resources/templates/subscription.html
```diff
@@ -251,32 +251,32 @@ <h3 class="modal-title text-center" id="formHeader">Subscription Form</h3>
...
                                     <input id="userNameInput" title="Enter user name" data-bind="textInput:userName" name="userName"
                                         placeholder="userName" class="form-control" type="text" />
                                     <span class="help-block"></span>
                                 </div>
 
-                                <label data-bind="visible: $root.showPassword()" class="pl-1 control-label font-weight-bold">Password*</label>
+                                <label data-bind="visible: $root.showPassword(), text: authenticationType() == 'BASIC_AUTH' ? 'Password*' : 'API-token'" class="pl-1 control-label font-weight-bold"></label>
                                 <div>
                                     <input id="passwordInput" title="Enter password" data-bind="visible: $root.showPassword() && ($root.mode() == 'edit' || $root.mode() == 'add'),textInput:$data.password"
                                         name="password" placeholder="password" class="form-control" type="password" />
```
J-Binieda-ericsson: Should this not change content to token when it change from Password* to API-token?

src/main/resources/templates/subscription.html
```diff
@@ -251,24 +251,32 @@ <h3 class="modal-title text-center" id="formHeader">Subscription Form</h3>
                                     </select>
                                 </div>
                             </div>
-                            <div class="p-1 border form-group" data-bind="visible: authenticationType() == 'BASIC_AUTH'">
+                            <div class="p-1 border form-group" data-bind="visible: authenticationType() == 'BASIC_AUTH' || authenticationType() == 'BASIC_AUTH_JENKINS' ">
```
Christoffer-Cortes: You might be able to use indexOf on an observable. I haven't tried it myself but maybe something like this 'authenticationType().indexOf('BASIC_AUTH') !== -1' would work?

PR #145, opened 2019-01-24
Dockerfile for building images for Docker and Kubernetes

tobiasake:
### Description of the Change
Adding a Dockerfile so user can build Docker image for Docker and Kubernetes.

### Benefits
Make it easy to build Eiffel-Intelligence Front-end docker images.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:


src/main/docker/Dockerfile
```diff
@@ -1,13 +1,13 @@
...
-EXPOSE 8092 8778 9779
 
-# Create image with existing war file. User need to execute 'mvn package -DskipTest' before 'docker build'
-COPY ./target/eiffel-intelligence-frontend-*.war /deployments/eiffel-intelligence-frontend.war
+ARG URL
+RUN echo Building Eiffel-Intelligence Front-end image based on artifact url: ${URL}
 
-CMD /deployments/run-java.sh
+# Create image with existing war file. User need to execute 'mvn package -DskipTest' before 'docker build'
+RUN ["rm", "-fr", "/usr/local/tomcat/webapps/ROOT"]
```
e-edling-ericsson: Same question here as in eiffel-intelligence backend.

wiki/markdown/docker.md
```diff
@@ -33,15 +33,15 @@ Now docker image has build with tag "eiffel-intelligence-frontend:0.1"
 To run the produced docker image on the local Docker host, execute this command: 
 
 
-`docker run -p 8034:8091 --expose 8091 -e server.port=8091 -e logging.level.root=DEBUG -e logging.level.org.springframework.web=DEBUG -e logging.level.com.ericsson.ei=DEBUG eiffel-intelligence-backend:0.1`
+`docker run -p 8071:8080 --expose 8080 -e server.port=8080 -e logging.level.log.level.root=DEBUG -e logging.level.org.springframework.web=DEBUG -e logging.level.com.ericsson.ei=DEBUG eiffel-intelligence-frontend:0.0.19`
```
e-pettersson-ericsson: Using this command the eiffel intelligence frontend won't know which backend to use, because no backend is specified. Add a flag which passes in the default backend instance to use.

src/main/docker/Dockerfile
tobiasake: This comes from early versions of @emichaf Dockerfiles,, don't know why we have it. Looks like to cleaning the default Tomcat webapp which is replaced by Eiffel-Intelligence Front-end app.
The difference of running the command a brackets "[]" and not, I think its something envrionment and java properties is treated differently. But in this case is should not matter which format that is used, since it is only a normal remove command that is executed.

wiki/markdown/docker.md
e-edling-ericsson: I think it should be clearly stated that those values are just examplevalues. Right now it seems like that line should be executed that way, You could otherwise write something like <EXPOSED_PORT>:<CONTAINER_PORT> or something like that for all the "example" values.

e-pettersson-ericsson: I agree with @e-edling-ericsson. It would also be easier to maintain the documentation if we have generic example values that don't have to be updated everytime we make a release... 
(thinking of the version tag used in the above command, this PR was written when EI was still 0.0.19 but 1.0.0 has already been released and it will keep changing in the future).

wiki/markdown/docker.md
```diff
@@ -57,17 +57,24 @@ It is possible to set all Spring available properties via docker environment "-e
 ## Docker flags
 
 
-<B>"--expose 8091"</B> - this Docker flag tells that containers internal port shall be exposed to outside of the Docker Host. This flag do not set which port that should be allocated outside Docker Host on the actual server/machine.
+<B>"--expose 8080"</B> - this Docker flag tells that containers internal port shall be exposed to outside of the Docker Host. This flag do not set which port that should be allocated outside Docker Host on the actual server/machine.
 
 
-<B>"-p 8034:8091"</B> - this Docker flag is mapping the containers external port 8034 to the internal exposed port 8091. Port 8034 will be allocated outside Docker host and user will be able to access the containers service via port 8034.
+<B>"-p 8071:8080"</B> - this Docker flag is mapping the containers external port 8034 to the internal exposed port 8091. Port 8034 will be allocated outside Docker host and user will be able to access the containers service via port 8034.
```
e-pettersson-ericsson: Update the text, it still says 8034 even though you've changed the example to 8071.

tobiasake: Done.

PR #161, opened 2019-02-08
Various subscription page fixes

Christoffer-Cortes:
### Description of the Change
- Bulk and single delete now use a bootstrap modal, which they share, instead of js modal.
- Bulk and single delete now share the same function for deletion.
- When uploading subscriptions from a file, removing these subscriptions and trying to add them again from the same file, it would not upload. This is no longer the case.
- Upload function is now much simplified.
- Checkboxes are now only usable when the backend is up and there is at least one subscription.
- Checkbox will now clear after a bulk delete and not remain ticked.
- Checkbox for bulk delete should now properly register click event even when you enter subscription page with backend being down and then going up.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/src/main/resources/static/js/subscription.js
+++ b/src/main/resources/static/js/subscription.js
@@ -110,6 +110,21 @@ jQuery(document).ready(function () {
         }
         $("#back_end_down_warning").toggle(!backendStatus);
         toggleButtonsDisabled(!backendStatus);
+        toggleCheckboxesDisabled(!backendStatus);
+    }
+
+    function toggleCheckboxesDisabled(disabled) {
+        $('#check-all').prop("disabled", disabled);
+        $('.data-check').prop("disabled", disabled);
+    }
+
+    function toggleButtonsDisabled(disabled) {
+        $('#addSubscription').prop("disabled", disabled);
+        $('#bulkDelete').prop("disabled", disabled);
+        $('#uploadSubscription').prop("disabled", disabled);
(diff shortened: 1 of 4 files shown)


src/main/resources/static/js/subscription.js
```diff
@@ -519,6 +547,38 @@ jQuery(document).ready(function () {
...
+            beforeSend: function () {
+            },
+            success: function (data, textStatus) {
+                reload_table();
+            },
+            error: function (XMLHttpRequest, textStatus, errorThrown) {
+                reload_table();
+                var responseJSON = JSON.parse(XMLHttpRequest.responseText);
+                for (var i = 0; i < responseJSON.length; i++) {
+                    window.logMessages("Error deleteing subscription: [" + responseJSON[i].subscription + "] Reason: [" + responseJSON[i].reason + "]");
```
e-pettersson-ericsson: deleteing --> deleting

src/main/resources/static/js/subscription.js
```diff
@@ -519,6 +547,38 @@ jQuery(document).ready(function () {
     // /Stop ## Reload Table#################################################
 
     // /Start ## Bulk delete#################################################
+    function deleteSubscription(subscriptionsToDeleteString) {
```
tobiasake: I guess one or several subscriptions can be deleted with this function,, then its plural and the function name should be in plural with a "s" character in the end of the name ->  "deleteSubscriptions(...)"

PR #185, opened 2019-02-25
Add success check in test execution flow

e-edling-ericsson:
### Description of the Change

This PR makes it possible to know if a flow is complete or not.
### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: 

### DIFF
--- a/src/systemtest/resources/features/ArtifactFlow.feature
+++ b/src/systemtest/resources/features/ArtifactFlow.feature
@@ -11,36 +11,35 @@ Feature: Artifact system test
     And a jenkins job "TCSTCFJob" from "src/systemtest/resources/JenkinsShellScripts/TCSTCFScript.txt" is created with parameters: EVENT_ID,
     And a jenkins job "CLMJob" from "src/systemtest/resources/JenkinsShellScripts/CLMScript.txt" is created with parameters: EVENT_ID,
     And a jenkins job "ArtPJob" from "src/systemtest/resources/JenkinsShellScripts/ArtPScript.txt" is created with parameters: EVENT_ID,
-    And a jenkins job "FlowCompleteJob" from "src/systemtest/resources/JenkinsShellScripts/FlowCompleteScript.txt" is created with parameters: EVENT_ID,
+    And a jenkins job "FlowCompleteJob" from "src/systemtest/resources/JenkinsShellScripts/FlowCompleteScript.txt" is created with parameters: ,
 
     #####Add CLMSubscription to EI#####
-    Given subscription object "CLMSubscription" is created which will trigger "CLMJob"
+    Given subscription object "CLMSubscription" is created which will trigger "CLMJob" with parameters
     When notification with key "EVENT_ID" and value "id" is added to "CLMSubscription"
     And condition with jmespath "gav.artifactId=='ArtC2'" is added to "CLMSubscription"
     Then we send the "CLMSubscription" to eiffel intelligence for creation.
 
     #####Add TCTSubscription to EI#####
-    Given subscription object "TCTSubscription" is created which will trigger "TCTJob"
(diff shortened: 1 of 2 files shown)


src/systemtest/resources/features/ArtifactFlow.feature
```diff
@@ -11,36 +11,35 @@ Feature: Artifact system test
     And a jenkins job "TCSTCFJob" from "src/systemtest/resources/JenkinsShellScripts/TCSTCFScript.txt" is created with parameters: EVENT_ID,
     And a jenkins job "CLMJob" from "src/systemtest/resources/JenkinsShellScripts/CLMScript.txt" is created with parameters: EVENT_ID,
     And a jenkins job "ArtPJob" from "src/systemtest/resources/JenkinsShellScripts/ArtPScript.txt" is created with parameters: EVENT_ID,
-    And a jenkins job "FlowCompleteJob" from "src/systemtest/resources/JenkinsShellScripts/FlowCompleteScript.txt" is created with parameters: EVENT_ID,
+    And a jenkins job "FlowCompleteJob" from "src/systemtest/resources/JenkinsShellScripts/FlowCompleteScript.txt" is created with parameters: ,
```
a-breid-ericsson: Gurka statements should not end with , or .?

src/systemtest/resources/features/ArtifactFlow.feature
```diff
@@ -11,36 +11,35 @@ Feature: Artifact system test
...
 
     #####Add CLMSubscription to EI#####
-    Given subscription object "CLMSubscription" is created which will trigger "CLMJob"
+    Given subscription object "CLMSubscription" is created which will trigger "CLMJob" with parameters
```
a-breid-ericsson: Is parameters added in this step? It looks like the parameters is added in next step..

src/systemtest/resources/features/ArtifactFlow.feature
e-edling-ericsson: Actually this has to. It's a workaround to use lists on one line. The empty lines need some text after ":" and "," is the seperator which means it won't be interpreted as a parameter. I can remove them from the lines with parameters but i'd rather keep them on all lines in that case.

src/systemtest/resources/features/ArtifactFlow.feature
e-edling-ericsson: When the subscription is created, it needs to know if it will carry parameters or not since the notificationmeta will be different e.g "/build" or "/buildWithParameters". I should maybe change the name to, "which will carry parameters"?

a-breid-ericsson: I would rather see that notification meta is set on it´s own line. But since the gurka is created with a lot of magic behind then I presume it must be ok cause it would require too much change in already merged gurka-code.

src/systemtest/resources/features/ArtifactFlow.feature
a-breid-ericsson: I would prefer that parameters are added in it´s own line, but since this is already merged I guess it would require too much changes to implement and this may be ok.

src/systemtest/resources/features/ArtifactFlow.feature
e-edling-ericsson: Ye i think as little details as possible should be written in cucumber code. If we can hide implementation details like that behind normal language it's great. It would be too cluttered to set all those fields here.

src/systemtest/resources/features/ArtifactFlow.feature
e-edling-ericsson: I think that was to cluttered and it did not work to send 0 parameters in that case.

PR #195, opened 2019-03-14
Add support to validate multiple emails

e-edling-ericsson:
### Description of the Change

- Frontend now supports validation of multiple emails
  in the notification meta.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: 

### DIFF
--- a/src/main/resources/static/js/subscription.js
+++ b/src/main/resources/static/js/subscription.js
@@ -839,9 +839,9 @@ jQuery(document).ready(function () {
             var emails = notificationMeta.split(",");
             for(var email of emails) {
                 email = email.trim();
-                var regExpression = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
-                var isInvalidEmailAddress = (!regExpression.test(email) && email != "");
-                if (isInvalidEmailAddress) {
+                var validEmailRegExpression = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+                var isValidEmailAddress = validEmailRegExpression.test(email);
+                if (!isValidEmailAddress) {
                     $('#invalidNotificationMeta').text(email + " not a valid email.");
                     error = true;
                     break;
--- a/src/main/resources/static/js/subscription.js
+++ b/src/main/resources/static/js/subscription.js
@@ -827,20 +827,25 @@ jQuery(document).ready(function () {
         $('#notificationMeta').removeClass("is-invalid");
         $('#errorExists').hide();
 
-        var notoficationMetaIsEmpty = (
+        var notificationMetaIsEmpty = (
             !allowEmpty && notificationMeta == "" || !allowEmpty && notificationMeta.replace(/\s/g, "") == '""');
-        if (notoficationMetaIsEmpty) {
-            $('#invalidNotificationMeta').text("NotificationMeta must not be empty");
+        if (notificationMetaIsEmpty) {
+            $('#invalidNotificationMeta').text("NotificationMeta must not be empty.");
             error = true;
         } else if (vm.restPost()) {
             //validate url not implemented yet.
         } else if (!vm.restPost()) {
             // Validate email
-            var regExpression = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
-            var isInvalidEmailAddress = (!regExpression.test(notificationMeta) && notificationMeta != "");
-            if (isInvalidEmailAddress) {
-                $('#invalidNotificationMeta').text("Not a valid email.");
-                error = true;
+            var emails = notificationMeta.split(",");
+            for(var email of emails) {
+                email = email.trim();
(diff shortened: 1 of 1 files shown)


src/main/resources/static/js/subscription.js
```diff
@@ -827,20 +827,25 @@ jQuery(document).ready(function () {
...
             //validate url not implemented yet.
         } else if (!vm.restPost()) {
             // Validate email
-            var regExpression = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
-            var isInvalidEmailAddress = (!regExpression.test(notificationMeta) && notificationMeta != "");
-            if (isInvalidEmailAddress) {
-                $('#invalidNotificationMeta').text("Not a valid email.");
-                error = true;
+            var emails = notificationMeta.split(",");
+            for(var email of emails) {
```
e-pettersson-ericsson: I don't see any tests for this. Add a unittest which tries with multiple emails

src/main/resources/static/js/subscription.js
```diff
@@ -827,20 +827,25 @@ jQuery(document).ready(function () {
...
+                email = email.trim();
+                var validEmailRegExpression = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+                var isValidEmailAddress = validEmailRegExpression.test(email);
```
tobiasake: Yes, please unit test this long regexp validation mail check.

e-edling-ericsson: This is JS code that was written before. There are no unittests for the JS code. We only have seleniumtests to test the frontend and i don't think we should use those tests for this. They are too slow.

I agree that we should have unittests for the JS code although we must create a framework/investigate how we should unittest the JS code in that case.

src/main/resources/static/js/subscription.js
e-edling-ericsson: I left a comment on Tobias comment :)

src/main/resources/static/js/subscription.js
e-pettersson-ericsson: Is this tested with EI backend? In backend there are some unit tests on the subscription validator, so myabe extend those with multiple emails?

tobiasake: Ok, I missed it was java script part.
Very long time ago we had some validation code in Javascript in Frontend. Due to it was hard to unit test in JS and we wanted collect all subscription validations in one place, we placed all this validation checks in backend and we put the validation check result in the response back to the EI Frontend web page without closing the Create Subscription window. Windows was only closed if all validation checks passed in backend and subscription was created succesful in mongodb.

e-edling-ericsson: @e-pettersson-ericsson Yes that should be tested. That doesn't have anything with this code to do however. Backend has it's own validation which should be tested as well as the email sending to multiple recipients.

tobiasake: Can't this put in the validator in backend?

e-edling-ericsson: What do you mean? :)

tobiasake: It was decided very long time back to not put any validation here,, instead put the validation in backend, due to several reasons:
- Unit tests can be created for the validation implementation.
- Validation in one place, we do not have duplicated validations in both.
- Since all validation is in backend, validation checks is performed if subscriotions is uploaded without frontend and directly to backend.
It was for sure some more reason why we decided to have the validation in backend.

e-edling-ericsson: It was also decided not too long ago that the GUI should give feedback to the user if things are wrong, instantly. Not after trying to create the subscription. 

The backend API also validates the subscription today. So ye, it's done in two places but it is necessary if we want instant validation feedback while writing the subscriptions as well as validation if someone adds the subscription via REST.

tobiasake: Ok,, its ok to make that decision, as long it don't validate the first decision.
We made the first decision, very long time back, for a reason.
Same decision was made for all subscription template file , etc. , that it should be fetched from backend, so we don't have alot of duplicated files everywhere.

tobiasake: If you/we have decided something like this, and you/we have decided to say OK to violate first decision, you should add same check in backend as well, that check same thing.

e-edling-ericsson: I haven't decided this nor did i implement this validation. As i said, i'm not adding any functionality here, this has been here and was approved months ago after the exactly same discussion we are having now. I'm just fixing a bug that according to my story, already worked in the backend but was broken in the frontend.

a-breid-ericsson: The front-end validates several things like subscription name and that email is correct and that required fields contain data. This is a proper way to give instant feedback to the users and should be required from any kind of gui. The validation also needs to be in the back-end due to the possibility to add subscriptions via curl or other applications. And there is already validation there and there is no need to argue about it here. It is understandable that one do not want to have to update several validations on changes in how subscriptions look, but that should not go over the useability of a gui as that decission long time ago seem to have done.

tobiasake: This is nothing that has be requested by me only, architects and users has been involved.
Users complained earlier about it was possible to create to create subscriptions via restapi directly on Backend, but failed via frontend due to different validations in frontend and backend.
This was one of the source to why we,together with architects, decided to remove all validation checks in frontend to only have it in backend. Have validation implementation in both frontend and backend is bad design. Then we are going back to the implementation we have before our decision. We made that decision so we can fullfill all requirements and have all validation implementation in one place so it same validation executed no matter which alternative the user chooses to create subscriptions and it will simplify the maintainance since we have the validation in one place. 
We had also complains about our subscription template files in frontend didn't follow correct format, issue to that we had duplicated files in both frontend and backend, these files was badly maintained and was not updated when subscription fomat changed. If it was updated, then files was updated only in backend.

Anyway,, it is really bad to have duplicate functionalities and files.

e-edling-ericsson: Then i think we should have a discussion/meeting whether we should keep the validations in frontend or not. Just because it was decided long time ago doesn't mean we can't decide something new if we think it is an improvement.

tobiasake: Decided that developer need to make sure backend make same validation check, or better check.
So I approve now.

PR #198, opened 2019-03-19
Update docker-compose for integration tests

e-pettersson-ericsson:
### Applicable Issues
During integration tests we use the docker-compose file in the Eiffel Intelligence backend repository to set up environment. Instead I moved the file to this repository, which can be reused for running system tests, when they are merged into master. This way we can remove the extra docker-compose file in the Eiffel Intelligence backend repositry in the root directory.

### Description of the Change
* Updated the existing docker-compose file to set up needed environment for both integration tests (and later also system tests) and updated configuration for newest EI release. 
* Updated the integration tests properties to use new rules in backend.
* Updated README file in src/main/docker/
* Added an env.bash file in src/main/docker/ to avoid cluttering docker-compose file with json and extracted the ports and images to make it easier to update a service in docker-compose
* Refactored the Travis build a little to use the local docker-compose file during integration tests

### Benefits
One docker-compose file per repository. Everything docker related in one place.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: Emelie Pettersson emelie.pettersson@ericsson.com


src/main/docker/docker-compose.yml
```diff
@@ -162,36 +353,41 @@ services:
       - spring.mail.password=
       - spring.mail.properties.mail.smtp.auth=false
       - spring.mail.properties.mail.smtp.starttls.enable=false
-      - er.url=eiffel-er:8082
+      - er.url=http://${HOST}:8084/search/
```
tobiasake: It hink you should use internal hostname on each service instead of HOST ip address.
By using HOST IP address the connection will exit docker host and enter the Docker host again. This is an issue if you are working from home and connected with VPN, the VPN tunnel will eat up the connection ER connection when it exits the dockerhost and it will never find the ER service. You will probably get a connection timeout or something in that situations. So better to use the Eiffel-ER internal Docker hostname / alias, e.g.: "eiffel-er" .

src/main/docker/docker-compose.yml
```diff
@@ -77,83 +77,274 @@ services:
...
-      - missedNotificationDataBaseName=eiffel2_intelligence_MissedNotification
+      - spring.data.mongodb.database=eiffel-intelligence-sourcechange
+      - missedNotificationDataBaseName=MissedNotification-sourcechange
+      - search.query.prefix=object
+      - aggregated.object.name=aggregatedObject
+      - spring.mail.properties.mail.smtp.auth=false
+      - spring.mail.properties.mail.smtp.starttls.enable=false
+      - er.url=http://${HOST}:8084/search/
+      - logging.level.root=OFF
+      - logging.level.org.springframework.web=ERROR
```
tobiasake: I think all logging besides root logging should be default set to DEBUG.

src/main/docker/docker-compose.yml
```diff
@@ -77,83 +77,274 @@ services:
...
+      - WAIT_DB_HOSTS=mongodb:27017
+
+  ############### Stop EI backend artifact rules ##############################
+
+
+  ############### Start EI backend sourcechange rules ##########################
+  ei-backend-sourcechange:
     restart: always
-    image: eiffelericsson/eiffel-intelligence-backend:1.0.0-1
+    image: eiffelericsson/eiffel-intelligence-backend:1.0.1
```
tobiasake: If its mandatory already now to execute env.bash before running docker-compose, then you could put all image versions in env.bash,, then we have all version collected in one place.

tobiasake: or you can put the whole image tag inclusive version in env.bash.

src/main/docker/docker-compose.yml
e-pettersson-ericsson: Fixed

src/main/docker/docker-compose.yml
e-pettersson-ericsson: I've moed out the ports to the services to env.bash, and I'll move out the image versions also

src/main/docker/docker-compose.yml
e-pettersson-ericsson: Fixed.

PR #203, opened 2019-04-25
Update to hyphened spelling of front-end and back-end

e-pettersson-ericsson:
### Applicable Issues
Different uses of spelling of the words back-end and front-end were used in documentation. Previously a mix of "backend", "back end" and "back-end" were used, confusing the reader. We previously agreed to use hyphened words when referring to the componens in the markdown documentation.

### Description of the Change
Updated the markdown documentation to use only hyphened words when referring to Eiffel Intelligence components, i.e. "back-end" and "front-end".

### Benefits
More consistent reading in GitHub pages.

### Alternative Designs
According to what I've found, traditional grammar rules state that "front end" refers to a noun while "front-end" is a compound adjective. But these definitions are not really applicable in all cases, and "frontend" har surfaced as an alternative (modern?) way of writing. [More info in this blog post](https://dzone.com/articles/blogging-programmers-style). I'm OK with using either spelling as long as we keep it consistent, and don't mix in documentation and code comments. What I would like is to use "frontend" and "backend" in code comments, to match our code examples and variable naming standards. 

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: Emelie Pettersson emelie.pettersson@ericsson.com

### DIFF
--- a/wiki/markdown/docker.md
+++ b/wiki/markdown/docker.md
@@ -90,8 +90,8 @@ This docker-compose file includes these components, [docker-compose.yml](https:/
 - Mail server
 - Jenkins
 - Eiffel REMReM services (Generate and Publish)
-- 3 instances of EI back-end (using different rule sets)
-- EI front-end (Using the local EI front-end Docker image build from previous steps)
+- 3 instances of EI backend (using different rule sets)
+- EI frontend (Using the local EI front-end Docker image build from previous steps)
 
 NOTE: Only MongoDB, RabbitMQ, ER and EI components are needed to start.
 The rest of the components can be commented out if not needed.
--- a/wiki/markdown/docker.md
+++ b/wiki/markdown/docker.md
@@ -1,7 +1,7 @@
 # Docker
 
-In Eiffel-Intelligence frontend source code repository, a Dockerfile is provided which helps the developer or user to build the local Eiffel-Intellegence frontend source code repository changes to a Docker image.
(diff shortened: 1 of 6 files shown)


e-pettersson-ericsson:
> Looks good ... Though I am not sure about the grammatical use of hyphen

Indeed, I'm not so sure either :) "front end" and "front-end" are OK according to what I've found, but on the other hand, "frontend" is the new trend, as it's easier to write. Compare with "e-mail" vs "email". I'm waiting for @a-breid-ericsson to review, as he had opinions about this also.

wiki/markdown/docker.md
```diff
@@ -87,10 +87,10 @@ This docker-compose file includes these components, [docker-compose.yml](https:/
 - MongoDb
 - RabbitMq
 - ER
-- EI-Backend
```
tobiasake: I think the component names can use "backend" and "frotend" format and when describing something in text we can use "back-end" and "front-end" format.

e-pettersson-ericsson: I'm thinking it might confuse readers, since this entire file is meant for documentation purposes. An alternative could be to refer to the actual names used in the docker-compose file, but are we sure these names won't change in the future? I'm sticking with this for now to follow the documentation style.

tobiasake: Since line above refer to docker-compose file, my mind is reading as how it is named in docker-compose file.
If the service name changes completely in docker-compose file, then documentation need to be updated as well.

PR #212, opened 2019-05-20
Fix ldap check

a-breid-ericsson:
### Applicable Issues
Front end not working in EI (Internet Explorer), sometimes displays subscription owner names even when ldap is disabled.

### Description of the Change
Solved problem with IE.
ldapEnabled is defined in one place and may be accessed using getter and setters.
Minor improvements and small fixes in code.

### Benefits
It works.

### Possible Drawbacks
People may still use IE

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:

### DIFF
--- a/src/main/resources/static/js/common.js
+++ b/src/main/resources/static/js/common.js
@@ -2,12 +2,30 @@ var router = new Navigo(null, true, '#');
 var frontendServiceUrl = $('#frontendServiceUrl').text();
 var frontendServiceBackEndPath = "/backend";
 var timerInterval;
+var ldapEnabled = true;
+
+// Start ## getters and setters
+
+function isLdapEnabled(){
+    return Boolean(ldapEnabled);
+}
+
+function setLdapEnabled(value){
+    ldapEnabled = Boolean(value);
+}
+
+// End   ## getters and setters
 
 function addBackendParameter(url) {
+    if (!String.prototype.includes) {
+        String.prototype.includes = function (str) {
+            return this.indexOf(str) !== -1;
+        };
+    }
     if (!sessionStorage.selectedActive) {
         return url;
     }
-    var delimiter = "";
+    var delimiter = "?";
     var parameterKey = "backendname";
 
     if (url.includes("?")) {
@@ -32,7 +50,7 @@ AjaxHttpSender.prototype.sendAjax = function (contextPath, type, data, callback,
     if (!dataType) {
         dataType = "json";
     }
-    url = addBackendParameter(frontendServiceUrl+contextPath)
+    url = addBackendParameter(frontendServiceUrl+contextPath);
     $.ajax({
         url: url,
         type: type,
@@ -61,7 +79,7 @@ AjaxHttpSender.prototype.sendAjax = function (contextPath, type, data, callback,
             }
         }
     });
-}
+};
 // /Stop ## Global AJAX Sender function ##################################
 
 function formatUrl(host, port, useHttps, contextPath) {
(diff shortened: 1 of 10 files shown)


src/main/resources/static/js/login.js
```diff
@@ -1,28 +1,13 @@
 jQuery(document).ready(function () {
+    function reroutToSubscritpion() {
+        checkBackendSecured();
+        // if (ldapEnabled == false || (ldapEnabled == true && currentUser != null)) {
+        if (isLdapEnabled() && currentUser != null && currentUser != undefined || !isLdapEnabled()) {
```
tobiasake: Why is last "!isLdapEnabled()" needed? It should be first in the if statement,, no need of doing the other checks before if isLdapEnabled is disabled.

src/main/resources/static/js/subscription.js
```diff
@@ -467,81 +458,74 @@ jQuery(document).ready(function () {
                     "render": function (data, type, row, meta) {
                         if (data == undefined || row == undefined) {
                             window.logMessages("Error: Subscription data is not defined");
-                            return ''
+                            return '';
                         }
-                        subscriptionOwner = row.ldapUserName
-                        subscriptionName = data.subscriptionName
+                        subscriptionOwner = row.ldapUserName;
```
tobiasake: What if ldapUsername don't exist in Subscription?

a-breid-ericsson: subscriptionOwner will be undefined, there is a check for that when determing if owner exist for subscription.

src/main/resources/static/js/login.js
a-breid-ericsson: Done. It should only be !isLdapEnabled() cause if ldap is disabled then we should not show login page, and if ldap is not disabled then it must of course be true, så the other check for ldap is not needed.

src/main/resources/static/js/login.js
```diff
@@ -1,28 +1,12 @@
 jQuery(document).ready(function () {
+    function reroutToSubscritpion() {
```
Christoffer-Cortes: rerouteToSubscription

src/main/resources/static/js/login.js
evasiba-ericsson: Please group the logical expression used here. I assume that isLdapEnabled() && currentUser != null && currentUser != undefined  is one group and  !isLdapEnabled() is the second group. If that is the case add parathesis around the second group or make it a variable.

The expression is confusing at the first glance and it is frustrating to require to much attention to understand it.

a-breid-ericsson: Done, reused code already written for other function.

src/main/resources/static/js/login.js
a-breid-ericsson: Done

src/main/resources/static/js/common.js
```diff
@@ -2,19 +2,40 @@ var router = new Navigo(null, true, '#');
...
         return url;
     }
-    var delimiter = "";
-    var parameterKey = "backendname";
 
-    if (url.includes("?")) {
+    // If String includes function does not exist (IE) add function.
+    if (!String.prototype.includes) {
+        String.prototype.includes = function (str) {
+            return this.indexOf(str) !== -1;
```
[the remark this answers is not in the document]
a-breid-ericsson: Done. Although I was considering naming it isSubstring, but I believe everyone knows in coding what needle in haystack means.

PR #214, opened 2019-05-23
Bug regex changes in front-end to make it similar regex in back-end

saif-ericsson:
### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:


src/main/resources/static/js/subscription.js
```diff
@@ -861,9 +861,15 @@ jQuery(document).ready(function () {
             $('#subscriptionNameInput').addClass("is-invalid");
             error = true;
         }
-
+        
+        // When this regExp need to be changed then remember to change the one in the
+        // back-end (invalidSubscriptionNameRegex in SubscriptionValidator.java), which do the same
+        // invalid subscription name check. The two
+        // regEx always need to be the same for ensuring the same check.
         // /(\W)/ Is a regex that matches anything that is not [A-Z,a-z,0-8] and _.
```
tobiasake: Shouldn't it be possible to use dash "-" in SubscriptionName also, or is it only underscore?

src/main/resources/static/js/subscription.js
```diff
@@ -896,6 +902,10 @@ jQuery(document).ready(function () {
             var emails = notificationMeta.split(",");
             for(var email of emails) {
                 email = email.trim();
+                // When this regExp need to be changed then remember to change the one in the
+                // back-end (validEmailRegExpression in SubscriptionValidator.java), which do the same
+                // email validation check. The two
+                // regEx always need to be the same for ensuring the same check.
```
tobiasake: Manual test some mail addresses that should succeed and some mail addresses that should fail before merge.

src/main/resources/static/js/subscription.js
saif-ericsson: This regEx doesnt accept dash (-) in the subscription name... as we discussed, it shouldnt be in the name.

tobiasake: Ok,,, I just trying to think of what and which subscription name variants that user want to use.

PR #215, opened 2019-05-28
Add status bar on login page, centralized status handling

a-breid-ericsson:
### Applicable Issues
No status bar on login.
Status handling spread on many files.

### Description of the Change
- A status bar may now be shown on the login page if back end goes
  down
- Status handling has been moved to its own file
- Functionality previously handled in ajax request has now been moved
  to its own functions
- Several new getters and setters

### Alternate Designs
-Status handling to be done in common 

### Benefits
Easier to find the code handling status in its own file.

### Possible Drawbacks
More files.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:

### DIFF
--- a/src/main/resources/static/js/login.js
+++ b/src/main/resources/static/js/login.js
@@ -66,4 +66,10 @@ jQuery(document).ready(function () {
     ko.cleanNode(observableObject);
     var model = new loginModel();
     ko.applyBindings(model, observableObject);
+
+    // Check EI Backend Server Status ########################################
+
+    checkBackendStatus();
+
+    // END OF EI Backend Server check #########################################
 });
--- a/src/main/resources/static/js/login.js
+++ b/src/main/resources/static/js/login.js
@@ -42,17 +42,20 @@ jQuery(document).ready(function () {
                 navigateToRoute('subscriptions');
             },
             error: function (XMLHttpRequest, textStatus, errorThrown) {
-                if (XMLHttpRequest.status == 401) {
-                    window.logMessages("Bad credentials");
+                var statusCode = XMLHttpRequest.status;
+
+                if (statusCode == 401) {
+                    window.logMessages("Error Status code: 401 'Unauthorized or bad credentials'");
                     $('#loginError').text("Invalid username and/or password!");
-                    $('#loginError').addClass("is-invalid");
-                    $('#loginError').show();
+                } else if (statusCode == 500 && !isBackEndStatusOk()) {
+                    window.logMessages("Error back-end is not reachable!");
+                    $('#loginError').text("Back-end might be unavailable!");
                 } else {
-                    window.logMessages("Unknown login error");
-                    $('#loginError').text("Unknown login error!");
-                    $('#loginError').addClass("is-invalid");
-                    $('#loginError').show();
+                    window.logMessages("Error Status Code: " + statusCode + " 'Unknown server error'");
                 }
+
+                $('#loginError').addClass("is-invalid");
+                $('#loginError').show();
             },
             complete: function () {
             }
--- a/src/main/resources/static/js/status-handler.js
+++ b/src/main/resources/static/js/status-handler.js
@@ -15,6 +15,11 @@ function checkBackendStatus() {
         return;
     }
 
+    if (!isBackEndStatusUpdated()) {
+        addStatusIndicator(statusType.WARNING, statusText.UNKNOWN_BACK_END_STATUS);
+        return;
+    }
+
     if (isBackEndStatusOk()) {
         removeStatusIndicator();
     } else {
--- a/src/main/resources/static/js/status-handler.js
+++ b/src/main/resources/static/js/status-handler.js
@@ -5,57 +5,12 @@ enable or disable different kinds of status icons/fields.
 It may also fore page refresh pages, remove subscriptions from subscription list
 and reload subscriptions in subscription list depending on the back-end status.
 */
-var pagesForStatusIndication = "subscriptions test-rules ei-info login";
-
-var statusType = {
-    success: "alert-success",
-    info: "alert-info",
-    warning: "alert-warning",
-    danger: "alert-danger"
-};
-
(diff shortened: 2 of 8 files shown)


src/main/resources/static/js/status-handler.js
```diff
@@ -0,0 +1,159 @@
+/*
+This js file intend to keep the status information about the current selected back-end
+and depending weather the back end status has changed and or is online or not will
```
tobiasake: You are mixing the words "back-end" and "back end",, decide one format.

src/main/resources/static/js/status-handler.js
```diff
@@ -0,0 +1,159 @@
...
+and reload subscriptions in subscription list depending on the back-end status.
+*/
+var pagesForStatusIndication = "subscriptions test-rules ei-info login";
+
+var statusType = {
+    success: "alert-success",
+    info: "alert-info",
+    warning: "alert-warning",
+    danger: "alert-danger"
+};
```
e-edling-ericsson: These are constants right? I would love to see them uppercase. Also, maybe we would even have a file with all sorts of constants?

src/main/resources/static/js/status-handler.js
```diff
@@ -0,0 +1,159 @@
...
+
+var statusText = {
+    backend_down: "<strong>Back end is down!</strong> Wait for it go up or switch to another back end before continuing!",
+    test_rules_disabled: "<strong>Test Rule service is disabled!</strong> To enable it set the backend property [testaggregated.enabled] as [true]"
+};
+
+var backEndStatus = true;
+var previousBackEndStatus;
```
e-edling-ericsson: I don't think you should change them to backEnd(from backend). Didn't we agree that we will merge the words as it is now accepted as an technical term just like "e-mail -> email", "plug-in" -> plugin etc.

src/main/resources/static/js/subscription.js
```diff
@@ -371,7 +304,7 @@ jQuery(document).ready(function () {
             "searching": true,
             // Load data for the table's content from an Ajax source
             "ajax": {
-                "url": addBackendParameter(frontendServiceUrl + "/subscriptions"),
+                "url": addBackendParameter(getFrontEndServiceUrl() + "/subscriptions"),
```
e-edling-ericsson: Is there any better way in JS to build the URL instead of "+"?

src/main/resources/static/js/subscription.js
```diff
@@ -585,7 +518,7 @@ jQuery(document).ready(function () {
 
     function getTemplate() {
         var req = new XMLHttpRequest();
-        req.open("GET", frontendServiceUrl + '/download/subscriptionsTemplate', true);
+        req.open("GET", getFrontEndServiceUrl() + '/download/subscriptionsTemplate', true);
```
e-edling-ericsson: What is "true" in this case? I think you should create a variable above so that it's easy to see what true in this case actually means :)

a-breid-ericsson: it is a "get" not "is" so it should return a variable that is not a Boolean.

src/main/resources/static/js/subscription.js
a-breid-ericsson: Did not find any better way to build strings, There is String("h" + "e" + "j") but all it does is to ensure what is built is a String type.

src/main/resources/static/js/status-handler.js
a-breid-ericsson: In documentation all got changed to back-end.

src/main/resources/static/js/status-handler.js
a-breid-ericsson: In Java uppercase would be prefered, in javascript this is debated.

src/main/resources/static/js/status-handler.js
a-breid-ericsson: Done.

src/main/resources/static/js/status-handler.js
e-edling-ericsson: Then why do you call it "back end" here then? With that logic you could as well use the name that is more commonly used as a technical term "backend". :) I won't put -1 because of this, it's up to you. I've just given my opinion.

src/main/resources/static/js/subscription.js
e-edling-ericsson: Maybe we should build our own "buildUrl" method that we can use then. What i'm worried about is all the trailing "/" that can make things complex. The "buildUrl" method should be able to build "www.something.com/" and "/foo/bar" to "www.something.com/foo/bar" and not "www.something.com//foo/bar".

src/main/resources/static/js/subscription.js
e-edling-ericsson: Yes, but i don't know what is sent in to that method without reading the docs for req.open. I think you should create the variable above with a describing name so it's easy to know whats happening.

src/main/resources/static/js/status-handler.js
e-edling-ericsson: okay 👍

src/main/resources/static/js/subscription.js
Christoffer-Cortes: This can be achieved with a regex but perhaps make another story for this where every ajax url gets checked in a common url method.

src/main/resources/static/js/subscription.js
a-breid-ericsson: there is a function called forlatUrl, this validates and builds the correct url already before it is stored in the session storage. This could of course be extended to better validation.

src/main/resources/static/js/login.js
```diff
@@ -66,4 +66,10 @@ jQuery(document).ready(function () {
     ko.cleanNode(observableObject);
```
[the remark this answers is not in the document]
a-breid-ericsson: Message regaring failure to login has been updated.

PR #216, opened 2019-06-07
Regex from eiffel-commons to validate subscription

saif-ericsson:
We need to validate subscription in both front-end and back-end. Currently, we are using regular expressions which are part of code in both front-end and back-end. Thus, if we need to change regular expressions, we need to change at both places. It is difficult to maintain. Therefore, it is decided that we create a library (for regex) at eiffel-commons and both front-end and back-end will extract it from there.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:


src/test/java/com/ericsson/ei/frontend/RegExProviderUtilsTest.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.ericsson.ei.frontend.utils.RegExProviderUtils;
+
+public class RegExProviderUtilsTest {
+
+    @Test
+    public void test() {
+        String regExForInvalidName = RegExProviderUtils.getRegEx("invalidName");
+        String regExForValidEmail = RegExProviderUtils.getRegEx("validEmail");
+        assertEquals(regExForInvalidName, "(\\W)");
```
a-breid-ericsson: Please remove the regexps here and instead validate that it is not empty, since if regexp changes then we have to change here.
so assertFalse(fetchedRegexp.isEmpty())

saif-ericsson: done

src/main/java/com/ericsson/ei/frontend/WebController.java
```diff
@@ -42,6 +44,8 @@ public String greeting(Model model) {
     @RequestMapping("/subscriptionpage.html")
     public String subscription(Model model) {
         model.addAttribute("frontendServiceUrl", frontEndUtils.getFrontEndServiceUrl());
+        model.addAttribute("invalidName", RegExProviderUtils.getRegEx("invalidName"));
+        model.addAttribute("validEmail", RegExProviderUtils.getRegEx("validEmail"));
```
tobiasake: Check my review comments in backend pull requests.

a-breid-ericsson: Took me a while to understand your comment, but then validName could be subscriptionName and validEmail could be notificationMeta?

Although I like the suggestion, but notificationMeta could be url or email, currently we do not use regex for url, but if we add that, how would it validate both email and url?

src/main/java/com/ericsson/ei/frontend/utils/RegExProviderUtils.java
```diff
@@ -0,0 +1,24 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.ericsson.eiffelcommons.utils.RegExProvider;
+
+public class RegExProviderUtils {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(RegExProviderUtils.class);
+
+    public static String getRegEx(String key) {
```
e-edling-ericsson: It feels unnecessary to do logic regarding the retrieval of the regex here. The base lib should in that case return it as string if that is the wanted structure. Could then skip this method and where you want to use the regex you could do for example RegExProvider.getEmailRegEx() which will return a string directly. This will need to implement that function in the shared lib though.

saif-ericsson: New pull request implements functionality as per our team discussion

PR #218, opened 2019-06-18
Better indication that user must login to CRUD subscriptions

a-breid-ericsson:
### Applicable Issues
Hard do see and or understand that user must login if not logged in.
User do not know he is logged in just a green icon, but anyone may be logged in.

### Description of the Change
- Text added to notifications and Login icon, Login will be replaced
  with username when user is logged in
- CRUD buttons for subscriptions have a mouseover popup indicating
  that user must login if user is not logged in
- Icons on locked buttons will be a keylock when disabled

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:

### DIFF
--- a/src/main/resources/static/js/common.js
+++ b/src/main/resources/static/js/common.js
@@ -268,7 +268,7 @@ function cropUsername(username) {
         return username;
     }
 
-    var breakUsernameAtIndex = getUsernameMaxDisplayLength() - ellipses.leangth;
+    var breakUsernameAtIndex = getUsernameMaxDisplayLength() - ellipses.length;
     return username.substr(0, breakUsernameAtIndex) + ellipses;
 }
 
--- a/src/main/resources/static/js/common.js
+++ b/src/main/resources/static/js/common.js
@@ -239,25 +239,60 @@ function viewModel(backendInstanceData) {
 // End ## Load Back end list ##
 
 // Start ## Login and Security ##
-function doIfUserLoggedIn(user) {
-    setCurrentUser(user);
+function doIfUserLoggedIn(username) {
+    setCurrentUser(username);
+    showLoggedInUserInformation(username);
+    unlockSubscriptionButtons();
+}
+
+function showLoggedInUserInformation(username) {
     $("#userItem").show();
-    $("#userItem").addClass("user-login");
-    $("#ldapUserName").text(user);
+    $("#login-nav-bar-icon").addClass("user-login-icon");
+    $("#login-nav-bar-text").text(cropUsername(username));
+    $("#ldapUserName").text(username);
     $("#loginBlock").hide();
     $("#logoutBlock").show();
+}
+
+function unlockSubscriptionButtons() {
+    $(".logged-out-buttons").addClass("hidden_by_default");
+    $(".logged-in-buttons").removeClass("hidden_by_default");
     $(".show_if_authorized").prop('disabled', false);
+    $(".show_if_authorized").prop('title', "");
+}
+
+function cropUsername(username) {
+    var usernameMaxLength = 15;
+
+    if (username.length <= usernameMaxLength) {
+        return username;
+    }
+
+    var breakUsernameAtIndex = usernameMaxLength - 3;
+    return username.substr(0, breakUsernameAtIndex) + "...";
 }
 
 function doIfUserLoggedOut() {
     setCurrentUser("");
+    showUserLoggedOutInformation();
+    lockSubscriptionButtons();
+    sessionStorage.setItem('errorsStore', []);
+}
+
+function showUserLoggedOutInformation() {
     $("#userItem").show();
-    $("#userItem").removeClass("user-login");
+    $("#login-nav-bar-icon").removeClass("user-login-icon");
+    $("#login-nav-bar-text").text("Login");
     $("#ldapUserName").text("Guest");
     $("#loginBlock").show();
     $("#logoutBlock").hide();
+}
+
+function lockSubscriptionButtons() {
+    $(".logged-out-buttons").removeClass("hidden_by_default");
+    $(".logged-in-buttons").addClass("hidden_by_default");
     $(".show_if_authorized").prop('disabled', true);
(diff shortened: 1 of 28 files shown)


a-breid-ericsson:
Tests are already implemented testing that buttons are disabled / enabled and login logout. Since it is not recommended to run too many selenium tests I say no to tests for such functions as reading username as it is not critical to the core functionality.

src/main/resources/static/js/common.js
```diff
@@ -239,25 +239,60 @@ function viewModel(backendInstanceData) {
...
+}
+
+function unlockSubscriptionButtons() {
+    $(".logged-out-buttons").addClass("hidden_by_default");
+    $(".logged-in-buttons").removeClass("hidden_by_default");
     $(".show_if_authorized").prop('disabled', false);
+    $(".show_if_authorized").prop('title', "");
+}
+
+function cropUsername(username) {
```
Christoffer-Cortes: Ellipses can be added with css using:
overflow-x: hidden;
text-overflow: ellipsis;
You can't control the amount of characters but it does add ellipses if the text goes outside your side width.

a-breid-ericsson: Yes, but it does not count how many characters has been given, it also requires a fixed width and it makes shorter names push notifications away.

src/main/resources/static/js/common.js
```diff
@@ -239,25 +239,60 @@ function viewModel(backendInstanceData) {
 // End ## Load Back end list ##
 
 // Start ## Login and Security ##
-function doIfUserLoggedIn(user) {
-    setCurrentUser(user);
+function doIfUserLoggedIn(username) {
```
e-edling-ericsson: Maybe we can change this name? doIfUserLogged in says that something happens when you log in but not what. If think it should be more like "setUserInformationAndUnlockSubscriptionButton". Yes, that is a long name since this method really does a lot of things :D Atleast we will know what that method does.

src/main/resources/static/js/common.js
```diff
@@ -239,25 +239,60 @@ function viewModel(backendInstanceData) {
...
+
+function unlockSubscriptionButtons() {
+    $(".logged-out-buttons").addClass("hidden_by_default");
+    $(".logged-in-buttons").removeClass("hidden_by_default");
     $(".show_if_authorized").prop('disabled', false);
+    $(".show_if_authorized").prop('title', "");
+}
+
+function cropUsername(username) {
+    var usernameMaxLength = 15;
```
e-edling-ericsson: Maybe we should set this constant somewhere else?

src/main/resources/static/js/common.js
```diff
@@ -239,25 +239,60 @@ function viewModel(backendInstanceData) {
...
+
+    if (username.length <= usernameMaxLength) {
+        return username;
+    }
+
+    var breakUsernameAtIndex = usernameMaxLength - 3;
```
e-edling-ericsson: same with this constant 3. Should be easier to change through some constant class/config or something.

src/main/resources/static/js/common.js
```diff
@@ -239,25 +239,60 @@ function viewModel(backendInstanceData) {
...
+    return username.substr(0, breakUsernameAtIndex) + "...";
 }
 
 function doIfUserLoggedOut() {
     setCurrentUser("");
+    showUserLoggedOutInformation();
+    lockSubscriptionButtons();
+    sessionStorage.setItem('errorsStore', []);
```
e-edling-ericsson: Same here, we should find a name that says what this method does. Or break out some method calls from this method entirely if it does too many things and put it one layer out(if that suits better). 

From only reading the name, there should be no "side effects"

src/main/resources/static/js/common.js
a-breid-ericsson: Calling multiple functions is not really good, since if one new function is added, then that function must be called from multiple places as all functions. But I agree that the naming is horrible.

setUserInformationAndUnlockSubscriptionButton() would lock this function to do only that. a better function name could be "executeAuthenticatedUserFunctions()" that wouyd indicate that in this function is several function regaridng authenticated user functionality.

src/main/resources/static/js/common.js
a-breid-ericsson: Done.

src/main/resources/static/js/common.js
a-breid-ericsson: Moved to function specific variable, should never change.

src/main/resources/static/js/common.js
a-breid-ericsson: Name changed.

src/main/resources/static/js/common.js
```diff
@@ -262,17 +262,18 @@ function unlockSubscriptionButtons() {
...
+    var ellipses = "...";
 
-    if (username.length <= usernameMaxLength) {
+    if (username.length <= getUsernameMaxDisplayLength()) {
         return username;
     }
 
-    var breakUsernameAtIndex = usernameMaxLength - 3;
-    return username.substr(0, breakUsernameAtIndex) + "...";
+    var breakUsernameAtIndex = getUsernameMaxDisplayLength() - numberOfDotsAfterUsernameOnTooLongName;
```
e-edling-ericsson: Maybe it could just use ellipses.length? :D

a-breid-ericsson: Done.

src/main/resources/static/js/common.js
```diff
@@ -239,28 +239,63 @@ function viewModel(backendInstanceData) {
...
 
-function doIfUserLoggedOut() {
+function cropUsername(username) {
+    var ellipses = "...";
+
+    if (username.length <= getUsernameMaxDisplayLength()) {
+        return username;
+    }
+
+    var breakUsernameAtIndex = getUsernameMaxDisplayLength() - ellipses.leangth;
```
Christoffer-Cortes: ellipsis and length are spellt wrong.

PR #221, opened 2019-07-04
Standardize error message handling

Christoffer-Cortes:
### Description of the Change
An attempt to standardize error response. 
- Error logging now has another method that takes the message value of a response.
- If an error callback is not defined the default is to try to print the message value.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/src/main/java/com/ericsson/ei/frontend/utils/BackEndInformationControllerUtils.java
+++ b/src/main/java/com/ericsson/ei/frontend/utils/BackEndInformationControllerUtils.java
@@ -60,8 +60,8 @@ public class BackEndInformationControllerUtils {
         } catch (Exception e) {
             LOG.error("ERROR!\n" + e.getMessage());
             return new ResponseEntity<>(
-                    "[{\"name\":\"Unable to load instances\",\"host\":\"NO HOST\",\"port\":\"NO PORT\",\"contextPath\":\"/\"}]",
-                    getHeaders(), HttpStatus.OK);
+                    "{\"message\": \"Failure when trying to load backend instances\"}",
+                    getHeaders(), HttpStatus.NOT_FOUND);
         }
     }
 
--- a/src/main/resources/static/js/login.js
+++ b/src/main/resources/static/js/login.js
@@ -20,7 +20,7 @@ jQuery(document).ready(function () {
         this.login = function (userState, remember) {
             var dataJSON = JSON.parse(ko.toJSON(userState));
             if (dataJSON.ldapUserName == "" || dataJSON.password == "") {
-                window.logMessages("Username and password fields cannot be empty");
+                logMessage("Username and password fields cannot be empty");
             } else {
                 var token = window.btoa(dataJSON.ldapUserName + ":" + dataJSON.password);
                 sendLoginRequest(token);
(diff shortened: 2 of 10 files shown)


src/main/java/com/ericsson/ei/frontend/utils/BackEndInformationControllerUtils.java
```diff
@@ -60,8 +60,8 @@
         } catch (Exception e) {
             LOG.error("ERROR!\n" + e.getMessage());
             return new ResponseEntity<>(
-                    "[{\"name\":\"Unable to load instances\",\"host\":\"NO HOST\",\"port\":\"NO PORT\",\"contextPath\":\"/\"}]",
-                    getHeaders(), HttpStatus.OK);
```
tobiasake: isn't '[' ']' required in a JSON object in some Web-Browsers and in JS Ajax response?. I remember we added those to solve the errors we got in some web-browsers.

Christoffer-Cortes: Not that I have seen. In this case though it was used a dummy false backend instance but it also never reached this part. I've been unsuccessful in even trying to get anything printed when the backend list is wrong. The application will just shutdown.

src/main/resources/static/js/login.js
```diff
@@ -44,14 +44,15 @@ jQuery(document).ready(function () {
             error: function (XMLHttpRequest, textStatus, errorThrown) {
                 var statusCode = XMLHttpRequest.status;
 
+                // These error messages are not standardized.
                 if (statusCode == 401) {
-                    window.logMessages("Error Status code: 401 'Unauthorized or bad credentials'");
+                    logMessage("Error Status code: 401 'Unauthorized or bad credentials'");
```
e-pettersson-ericsson: Does all of these error messages have to start with 'Error' ? Isn't it enough with saying what went wrong?

src/main/resources/static/js/subscription.js
```diff
@@ -521,7 +521,7 @@ jQuery(document).ready(function () {
         request.responseType = "application/json;charset=utf-8";
         request.onload = function (event) {
             if (this.responseText == "") {
-                window.logMessages("Failed to download template, Error: Could not contact the backend server.");
+                logMessage("Failed to download template, Error: Could not contact the backend server.");
```
e-pettersson-ericsson: I would prefer to rewrite this sentence to: "Failed to download template. Could not contact the backend server." Since the jGrowl is already showing this as an error to the user.

src/main/resources/static/js/subscription.js
```diff
@@ -389,7 +389,7 @@ jQuery(document).ready(function () {
                     "data": null,
                     "render": function (data, type, row, meta) {
                         if (data == undefined || row == undefined) {
-                            window.logMessages("Error: Subscription data is not defined");
+                            logMessage("Error: Subscription data is not defined");
```
e-pettersson-ericsson: Keep the log messages consistent and remove the Error in the beginning of the sentence.

src/main/resources/static/js/testrules.js
```diff
@@ -16,7 +16,7 @@ jQuery(document).ready(function () {
             var context = ko.contextFor(event.target);
             self.rulesBindingList.splice(context.$index(), 1);
             if (self.rulesBindingList().length == 0) {
-                logMessages("Deleted all rule types, but we need atleast one Rule type, Here add default rule type");
+                logMessage("Deleted all rule types, but we need atleast one Rule type, Here add default rule type");
```
e-pettersson-ericsson: Forgot space between 'at least'. And the entire sentence reads a bit confusing ... Maybe something like: "All rule types were deleted, but we need at least one so we add a default rule here." would be better?

src/main/resources/static/js/login.js
Christoffer-Cortes: Probably not, it's kind of implied already. I'll fix this.

src/main/resources/static/js/subscription.js
Christoffer-Cortes: Done.

src/main/resources/static/js/subscription.js
Christoffer-Cortes: Done.

src/main/resources/static/js/testrules.js
Christoffer-Cortes: Done.

PR #228, opened 2019-07-12
Add page that displays rules from GET /rules endpoint

Christoffer-Cortes:
### Description of the Change
- Add rules page

### Sign-off

    indicated in the file; or

    in the file; or

    it.


src/main/resources/static/js/ei-rules.js
```diff
@@ -0,0 +1,10 @@
+jQuery(document).ready(function () {
+    var callback = {
+        success: function (responseData, textStatus) {
```
tobiasake: error callback missing and log the error to the UI?

Christoffer-Cortes: It's now default behavior that if no callback error is defined it will try to parse the response as an error message.

src/main/resources/templates/rules.html
```diff
@@ -0,0 +1,8 @@
+<div class="content">
+    <div class="col-12">
+        <h1 class="text-center">Active Rules File</h1>
```
e-pettersson-ericsson: I would like the rules.path to be displayed on this page, together with the rules. Also, since URL can also be used maybe file is not the best header? How about only "Active rules"?

Christoffer-Cortes: Done.

PR #234, opened 2019-08-06
Add script to run system tests

e-pettersson-ericsson:
### Applicable Issues
It is complicated to set up the proper environment and then run system tests with the right configurations. A systemtest script would help minimize the trouble to regularly run the system tests.

### Description of the Change
* Added a script which makes it easier to start up the proper environment with the latest code changes in EI back-end and front-end and then perform system tests on these. 
**NOTE**: Still working on having different verbosity levels for the script ... but feel free to review already now!
* Updated the system tests to use CSRF authentication towards Jenkins.

* **Updates in docker-compose file allows us to build and use docker images from latest code changes for EI and Jenkins.**
In our system tests we use JenkinsManager from eiffel-commons to create Jenkins jobs. At the same time we also install missing Groovy plugin in Jenkins. The problem with that was that Jenkins needed to be restarted after installation, and before it was up again the tests steps continued ... and failed to get a  response from Jenkins. 
I'm installing the Groovy plugin directly in the Jenkins container we start up for the environment. With this change the groovy plugin is instead installed directly in the Docker container from the beginning, and this reduces the need of the tests having to wait for Jenkins to restart.

By adding the '--build' flag to the 'docker-compose up' command images for Eiffel Intelligence images will always be built based on the latest code changes. It is still possible to use specific versions of docker images but these will have to be downloaded first (**docker-compose pull**) from DockerHub in order to avoid always building images for Eiffel Intelligence. 

_Parenthesis_: I've tried running even only one systemtest flow in Travis (see commit log) but they fail due to random hickups in the environment (that does not appear locally) or too low memory in the virtual machine Travis is running on...

### Benefits
Easier to run system tests on latest code changes for EI, and avoiding test failures because of Jenkins issues with installing Groovy plugin.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by:


src/systemtest/resources/systemtest.sh
```diff
@@ -0,0 +1,165 @@
...
+
+Usage: systemtest.sh COMMAND [OPTIONS]
+
+COMMANDS:
+    test :
+        First sets up Docker containers needed for system tests, then check Jenkins
+        container is healthy before starting the system tests defined for Eiffel
+        Intelligence, using maven verify command. After executing the tests, the
+        Docker containers are stopped and removed.
+    start :
```
tobiasake: "start" command can be a bit confusing when script is named systemtest.
User might/commonly interpret "systemtest start" command as starting the tests, which is "test" command in this case.
Maybe rename "start" to e.g. "env-start" or something,
and rename "test" to "start", since script name is already named with "test" in the name.

e-pettersson-ericsson: Hmm, I agree but maybe renaming the script file is the way to go then. Is "eiffel-intelligence.sh" too vague? The ommands start and test would have the correct meaning in that case. And both EI front-end and back-end with surrounding services are started so I'm reluctant to naming the script only "eiffel-intelligence-frontend"...

tobiasake: toolbox.sh, entry.sh, entrypoint.sh, setup.sh, is some more suggestions.

src/systemtest/resources/eiffel-intelligence.sh
```diff
@@ -77,30 +96,32 @@ function usage {
 Usage: systemtest.sh COMMAND [OPTIONS]
```
tobiasake: You renamed the script to "eiffel-intelligence.sh ",,, systemtest.sh is not correct.

e-pettersson-ericsson: I'll fix this in upcoming changes

src/main/docker/README.md
```diff
@@ -19,15 +19,29 @@ docker build -t eiffel-intelligence-frontend --build-arg URL=./${EIFFEL_WAR} -f
...
+The --build flag ensures the Docker images will be rebuilt if you have other 
+Docker images with the same name:tag lying around. Eiffel Intelligence back-ends
+and front-end and Jenkins containers will be built before starting up the environment. 
 
     source src/main/docker/env.bash
-    docker-compose -f src/main/docker/docker-compose.yml up -d
+    docker-compose -f src/main/docker/docker-compose.yml up -d --build
+
+**To run with older (specific) versions of Eiffel Intelligence pulled from DockerHub:**
+To avoid building any Eiffel Intelligence images that doesn't exist locally,
```
Christoffer-Cortes: "doesn't exist" should be "don't exist"

e-pettersson-ericsson: Will fix :)

src/systemtest/resources/eiffel-intelligence.sh
```diff
@@ -0,0 +1,194 @@
...
+
+function verbose {
+  if [[ "$verbose" -eq "1" ]]; then
+    echo -e $1
+  fi
+}
+
+function call {
+  verbose "Executing: '$@'"
+  $1
```
tobiasake: what is this doing? Isn't $1 only one word/command from the whole command?
Shouldn't "call" be name exec or something? Since it looks like this function executes a cli command.

src/systemtest/resources/eiffel-intelligence.sh
```diff
@@ -0,0 +1,194 @@
...
+        add this command before any other command. Example:
+        'eiffel-intelligence.sh build test'
+    only_test :
+        Assuming the environment is already set up, this only executes the
+        system tests with maven command.
+
+OPTIONS:
+    -v | --verbose  : Turn on verbose mode while executing. This echoes the shell
+                      commands before executing them.
+    -h | --help     : Shows this helps text.
```
tobiasake: "helps" ,, shouldn't it be just "help"

e-pettersson-ericsson: Fixed

src/systemtest/resources/eiffel-intelligence.sh
e-pettersson-ericsson: Changed the name of the function

PR #240, opened 2019-08-13
Change documentation information about repeat

e-edling-ericsson:
### Description of the Change

- Change the information of repeat handling to make it more clear about multiple requirements and that it doesn't affect the repeat functionality.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

Signed-off-by: 

### DIFF
--- a/wiki/markdown/add-subscription.md
+++ b/wiki/markdown/add-subscription.md
@@ -60,8 +60,8 @@ is disabled in Jenkins._
 
 **Repeat:** It is possible to enable repeat, e.g. whether same subscription
 should be re-triggered for new additions to the aggregated object. If disabled,
-the notification will be triggered only the first time when conditions are
-fulfilled.
+the notification will be triggered only the first time when requirements are
+fulfilled. It doesn't matter if you have multiple requirements, it will only be triggered once.
 
 **Requirement and Conditions:** It is used to add a requirement, which need to
 be fulfilled before a subscription is triggered. Requirement is added in the form


wiki/markdown/add-subscription.md
```diff
@@ -60,8 +60,8 @@ is disabled in Jenkins._
 
 **Repeat:** It is possible to enable repeat, e.g. whether same subscription
 should be re-triggered for new additions to the aggregated object. If disabled,
-the notification will be triggered only the first time when conditions are
-fulfilled.
+the notification will be triggered only the first time when requirements are
+fulfilled. It doesn't matter if you have multiple requirements, it will only be triggered once.
```
tobiasake: I think thats not true. I think its per requirement we store in RepeatHandleDb. Which means if subscription has several requirements, and one requirement has matched, subscription can match two more times, one per requirement, if I remember correct. But code/implementation might have changed since I worked with it.

e-edling-ericsson: If the subscription has repeat flag set and it has not produced a notification, once all conditions has been fulfilled, it will add the matched aggregated object to repeatDBHandler. If a aggregated object ID exists in the repeatDBHandler and has repeat flag set to false, it won't produce any more notifications no matter how many requirements there are. At least how i interprets the code.

And it also makes sense, if I have repeat set to false, i don't want notifications to happen multiple times. The requirements will only make it possible to trigger a notification if requirement A or requirement B is fulfilled. It does/should not say that the notification should be trigger once for year "or" statement, that would be kind of wierd. At least in my eyes 😄

tobiasake: Ok,, great that you checked functionality,, then its ok :)

PR #246, opened 2019-08-20
Show JSON information inside pre blocks with pretty format

Christoffer-Cortes:
### Applicable Issues
Related to https://github.com/eiffel-community/eiffel-intelligence/pull/315

### Description of the Change
Information that can be parsed into JSON will now be inside a pre block to display it in pretty format.
Strings that are parsed into numbers and booleans are excluded otherwise json parser will treat these as json.

### Sign-off

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/src/main/resources/static/js/common.js
+++ b/src/main/resources/static/js/common.js
@@ -106,7 +106,9 @@ function isStringDefined(value) {
 
 function parseJSON(value) {
     try {
-        if (typeof value === 'number' || value.match(/^\d+$/) || value.match(/^(true|false)$/)) {
+        var isNumber = "/^\d+$/";
+        var isBoolean = "/^(true|false)$/";
+        if (typeof value === 'number' || value.match(isNumber) || value.match(isBoolean)) {
             return undefined;
         }
         return JSON.parse(value);
--- a/src/main/resources/static/js/common.js
+++ b/src/main/resources/static/js/common.js
@@ -106,12 +106,12 @@ function isStringDefined(value) {
 
 function parseJSON(value) {
     try {
-        var isNumber = "/^\d+$/";
-        var isBoolean = "/^(true|false)$/";
-        if (typeof value === 'number' || value.match(isNumber) || value.match(isBoolean)) {
+        var json = JSON.parse(value);
+        if (json !== null && typeof json === "object") {
+            return json;
+        } else {
             return undefined;
         }
-        return JSON.parse(value);
     } catch (e) {
         return undefined;
     }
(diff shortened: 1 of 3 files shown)


src/main/resources/static/js/common.js
```diff
@@ -104,6 +104,17 @@ function isStringDefined(value) {
     return isDefined;
 }
 
+function parseJSON(value) {
```
tobiasake: What does this function do? What type of value? is it hte JsonObject.
Write a comment of what the function do or what it is used for.

Christoffer-Cortes: Any value. It's not a json object and not of a defined type. It checks if the value is a number or true/false, if not then it tries to parse the value into json. I'm not sure this requires a specific comment to understand to be honest. I could put the regex into variables but it should be quite easy to see what it does.

tobiasake: The issue is that when I look at the code I see only this function.
and it says only "parseJson" and Json in my world is a object "{ ... , ... }" with key and values. It can be any value . It doesn't say anythong what true, false, number, is used for and is nothing that can be derived from JSON name or JSON definition. Or am I missing something.
This is the background to my previous comment.

Christoffer-Cortes: Then we are in agreement, hahaha. If you read the PR description I try to explain why there is a check. The built in json parser treats numbers and booleans as json... so the check is there to make sure they are not positively validated and parsed. I only want things inside brackets to be validated and parsed.

tobiasake: ok,, but a developer that do look at the code later do not have the PR info. So the code must be explainable enough so developer understand the code.

Christoffer-Cortes: I can put that piece of code in a function called exlcudeNumbersAndBooleans or something but the problem is in JSON.parse method. Maybe put a link to the documentation of this function? I'm not fully sure myself what it allows or doesn't allow. It's a strange function though.

tobiasake: that function name says more than parseJson, since parseJson says more like a funtion that parse a Json object as string and return a Json object.
But this is my opinion, if other think I am thinking wrong and it should be treated differently, than you can skip my comment.

Christoffer-Cortes: I tried to find a regex to validate the json instead but I couldn't find any good expressions for javascript. There was an extremely accurate one in pcre but it doesn't work in javascript.

Christoffer-Cortes: I changed to parsing logic now. Maybe it makes more sense now?

tobiasake: ok.

